PR #2, opened 2025-05-14
Lower mlir.llvm to neura

tancheng:
Initial lower mlir.llvm.fadd to neura.add
 - Test is added
 - Can smoothly work with --insert-mov
 
TODO:
 - [x] Leverage .tb to perform the every pair of op lowering.
   - Done in https://github.com/coredac/dataflow/pull/3


tools/mlir-neura-opt/mlir-neura-opt.cpp
```diff
@@ -15,10 +17,14 @@ int main(int argc, char **argv) {
   registry.insert<mlir::neura::NeuraDialect>();
   registry.insert<mlir::func::FuncDialect>();
   registry.insert<mlir::arith::ArithDialect>();
+  registry.insert<mlir::LLVM::LLVMDialect>();
 
   mlir::registerPass([]() -> std::unique_ptr<mlir::Pass> {
     return mlir::neura::createLowerArithToNeuraPass();
   });
+  mlir::registerPass([]() -> std::unique_ptr<mlir::Pass> {
+    return mlir::neura::createLowerLlvmToNeuraPass();
```
ShangkunLi, line 26: I think it is more organized to define passes using the `.td` file since you may need to define many passes later. You can refer to the graph/kernel dialect for reference.

[NeuraDataflow/include/GraphDialect/GraphPasses.td](https://github.com/ShangkunLi/NeuraCompiler/blob/master/NeuraDataflow/include/GraphDialect/GraphPasses.td)

Just a minor suggestion ^_^, it's up to you to take it or not.

tancheng: Here are the pass registration. They are **declared** in the `include/` folder.

In your case, you also **define** them in https://github.com/ShangkunLi/NeuraCompiler/blob/master/NeuraDataflow/lib/GraphDialect/Transforms/GraphLinalgToGraph.cpp, same engineering effort imo?

ShangkunLi: MLIR enables us to use a `.td` file to manage our passes.

Just like https://github.com/llvm/llvm-project/blob/c54bc8bd07dc38135b6800922e261d573dcac956/mlir/include/mlir/Dialect/Arith/Transforms/Passes.td

You can also use the tablegen to define some command options using the `.td` easily.

tancheng: Ah I got it. Thanks for the pointers. Will refactor once there are more passes checked in.

ShangkunLi (approved):
Looks good to me.

PR #3, opened 2025-05-16
Pattern fusion template

tancheng:
A pattern fusion template
 - With example about fadd-fadd.
 - I decide **NOT** go forward with:
   - ~user providing any fusion pattern~
     - user needs to handle dependency, for large pattern, its complexity is same as writing up the pass/pattern_matcher
   - ~leverage tablegen~
     - Same complexity as writing up pass, and sometimes out of control and less flexibility, and user needs to think more/deep
 - I decide to provide specific flag for each type of fusion.


ShangkunLi (approved):
Everything looks good to me. ^_^

tancheng:
> Everything looks good to me. ^_^

Thanks, plz approve the dependent one :-)

Issue #5, opened 2025-05-17
[P1] An interpreter to run neura ops


tancheng:
Interpreter, simulator, or/and emulator to run neura operations, to verify the correctness.

tancheng:
@chz05 here is the existing interpreter/emulator: https://github.com/coredac/dataflow/blob/main/tools/neura-interpreter/neura-interpreter.cpp

Corresponding test: https://github.com/coredac/dataflow/tree/main/test/neura/interpreter

We can start with support interpreter for: https://github.com/coredac/dataflow/blob/main/test/neura/for_loop/test.mlir

PR #8, opened 2025-05-17
Implement a simple interpreter

tancheng:
Parses the neura op, and stores the result as float into a map.
 - first step towards https://github.com/coredac/dataflow/issues/5

TODO:
 - [ ] should support other value types, e.g., vector, tensor
 - [ ] should handle passed in arguments
 - [ ] support other neura operations and other dialects if necessary


ShangkunLi (approved):
Looks good to me.

Issue #10, opened 2025-05-17
[P1] Lower for loop


tancheng:
All ops from `for` loop (in llvm IR) should be lowered. Starting from simple dot product.

PR #14, opened 2025-05-18
Enable pass running on module

tancheng:
- Enable lower passes to work on `ModuleOp`
  - To make it be able to parse both llvm.func and mlir.func
  - First step towards https://github.com/coredac/dataflow/issues/10
- Introduce some other basic neura operations (vector floating add, i.e., vfadd) and fusion patterns (mul->add)

### DIFF
--- a/lib/Transforms/FusePatternsPass.cpp
+++ b/lib/Transforms/FusePatternsPass.cpp
@@ -8,30 +8,92 @@ using namespace mlir;
 
 namespace {
 
-struct FuseFAddFAddPattern : public RewritePattern {
-  FuseFAddFAddPattern(MLIRContext *ctx)
-      : RewritePattern("neura.fadd", /*benefit=*/1, ctx) {}
+struct FuseFAddFAddPattern : public OpRewritePattern<neura::FAddOp> {
+  using OpRewritePattern::OpRewritePattern;
 
-  LogicalResult matchAndRewrite(Operation *op, PatternRewriter &rewriter) const override {
-    auto first = dyn_cast<neura::FAddOp>(op);
-    if (!first || !first->hasOneUse()) return failure();
+  LogicalResult matchAndRewrite(neura::FAddOp second,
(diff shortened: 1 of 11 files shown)


lib/Transforms/FusePatternsPass.cpp
```diff
@@ -8,40 +8,116 @@ using namespace mlir;
...
     return success();
   }
 };
 
-struct FusePatternsPass : public PassWrapper<FusePatternsPass, OperationPass<func::FuncOp>> {
+struct FuseFMulFAddPattern : public OpRewritePattern<neura::FAddOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(neura::FAddOp add,
+                                PatternRewriter &rewriter) const override {
```
ShangkunLi, line 59: Based on the definition of `neura.fmul_fadd` operation:

```
def Neura_FMulFAddOp : Op<NeuraDialect, "fmul_fadd"> {
   let summary = "Fused fmul(fadd(a, b), c)";
   let arguments = (ins AnyFloat:$a, AnyFloat:$b, AnyFloat:$c);
   let results = (outs AnyFloat:$result);
   // let assemblyFormat = "$a `,` $b `,` $c attr-dict `:` type($result)";
   let traits = [SameOperandsAndResultElementType];
 }
```

The outer operation is `fmul`, so you pattern you should match here is `neura::FMulOp`.

Your implementation here corresponds to the `nuera.fadd_fmul` operation. That is `fadd(fmul(a, b), c)`, which is different from the definition of `nuera.fmul_fadd`.

tancheng: Good catch, thanks! Typo made by GPT. Update the `neura.fmul_fadd`'s `summary`:
```
let summary = "Fused fadd(fmul(a, b), c)";
```

yuqisun (approved):

PR #15, opened 2025-05-19
[fix] Typo in interpreter.mlir test

tancheng:


tancheng:
Not sure why still fails, but probably be fine once we merge this

yuqisun (approved):

Issue #18, opened 2025-05-22
[P0] Verify the clang++ version used in the test


tancheng:
Github action probably has a different clang version from my local, leading to different compilation behavior so tests are passed local but not on Github action.

This is from my local:
```
root@a354e4c22f79:/workspace/dataflow/test/neura/for_loop# clang++ -v
Ubuntu clang version 14.0.0-1ubuntu1.1
Target: aarch64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
Found candidate GCC installation: /usr/bin/../lib/gcc/aarch64-linux-gnu/11
Selected GCC installation: /usr/bin/../lib/gcc/aarch64-linux-gnu/11
Candidate multilib: .;@m64
Selected multilib: .;@m64
```

We can either update local or Github action. @yuqisun plz help check the Github action clang version.

yuqisun:
Hi @tancheng ,

Looks it costs more than 1 hour to install LLVM from source code, can we do this with pre-compiled pkg, same as [Mapper](https://github.com/yuqisun/CGRA-Mapper/blob/e0569b01b27a4072fca7f9ec41422fd7f3b832c5/.github/workflows/cmake.yml#L31) but version change to 14?

Thanks,

tancheng:
> Mapper

Plz help try, we need the path to the LLVM's library/header files to compile/build this project: https://github.com/coredac/dataflow/blob/ff914bc44d0c2cb5589ca962ff32cbe372ea9871/CMakeLists.txt#L29-L31

And now we are leveraging a specific commit of the LLVM: https://github.com/coredac/dataflow/blob/ff914bc44d0c2cb5589ca962ff32cbe372ea9871/README.md?plain=1#L10
I don't mind change the dependency though. Plz try a latest stable LLVM-dev, and we can port our project for that version. I think there shouldn't too many changes needed on our side.

PR #19, opened 2025-05-22
[P0] Verify the clang++ version used in the test #18

yuqisun:
Test to print LLVM version of Github Actions.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -59,6 +59,8 @@ jobs:
           -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
 
         cmake --build . --target check-mlir
+        clang++ -v
+        dpkg -l | grep llvm
         
     # setup mlir-cgra
     - name: setup dataflow tool-chain


tancheng:
Hi @yuqisun, so the apt-installed method works on your local machine/docker?

yuqisun:
> Hi @yuqisun, so the apt-installed method works on your local machine/docker?

Nope :(, revert just now, looks MLIR is not included in apt-install. Have to build by source code.

tancheng:
> > Hi @yuqisun, so the apt-installed method works on your local machine/docker?
> 
> Nope :(, revert just now, looks MLIR is not included in apt-install. Have to build by source code.

What about `sudo apt install llvm-14 clang-14 libclang-14-dev libmlir-14-dev mlir-14-tools`? Can try on your local first?

.github/docker/Dockerfile
tancheng, line 1: Why no llvm version specified in this docker file? default is which version?

.github/workflows/main.yml
```diff
@@ -34,13 +34,16 @@ jobs:
...
+    # install clang for LLVM build
+    - name: install clang
+      run: sudo apt-get install -y clang
+
     # setup LLVM
     - name: install a specific version of LLVM
       working-directory: ${{github.workspace}}
       run: |
-        git clone https://github.com/llvm/llvm-project.git
+        git clone --depth 1 --branch release/19.x https://github.com/llvm/llvm-project.git
```
tancheng, line 45: So this 19.x could work same as the commit `cd70802`? No error during our dialect's building?

.github/workflows/main.yml
```diff
@@ -55,10 +58,14 @@ jobs:
...
           -DMLIR_INSTALL_AGGREGATE_OBJECTS=ON \
           -DLLVM_ENABLE_RTTI=ON \
+          -DCMAKE_BUILD_WITH_INSTALL_RPATH=ON \
           -DCMAKE_C_COMPILER_LAUNCHER=ccache \
           -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
 
         cmake --build . --target check-mlir
+        dpkg -l | grep llvm
+        sudo apt-get purge llvm-13
+        sudo apt-get purge llvm-15
```
tancheng, line 68: Why we need to llvm? Why cannot this be aligned with llvm-19.x?

README.md
```diff
@@ -54,3 +54,16 @@ Build LLVM & Neura
  $ /workspace/llvm-project/build/bin/llvm-lit *
 ```
```
tancheng, line 56: Should we update `Check out to commit `cd70802` (a stable version randomly picked, will sync to the latest version).` to align with your changed `19.x`?

.github/docker/Dockerfile
yuqisun: release/19.x is used in Dockerfile when checkout source code:
`&& until git clone --depth 1 --branch release/19.x https://github.com/llvm/llvm-project.git; do sleep 5; done \`

.github/workflows/main.yml
yuqisun: This is to purge other versions of llvm in case version conflict issue.

README.md
yuqisun: Sure, `cd70802` is now the latest commit for `release/19.x` branch, let me add it also in case the `release/19.x` is not stable someday.

tancheng: I mean can we just do `git checkout release/19.x`? I randomly picked `cd70802`, I don't believe it is exactly the commit for `release/19.x`...

tancheng: Wow i swear i just randomly picked one.

Qq, why people would change the commits in a already release? I thought it is what it is forever.

yuqisun: Believe you are right, I was wrong (thought there might be continuous development on the release). So, shallow clone 19.x branch would be good.

README.md
```diff
@@ -25,6 +24,7 @@ Build LLVM & Neura
      -DLLVM_ENABLE_LLD=ON \
      -DMLIR_INSTALL_AGGREGATE_OBJECTS=ON \
      -DLLVM_ENABLE_RTTI=ON \
+     -DCMAKE_BUILD_WITH_INSTALL_RPATH=ON \
```
tancheng, line 27: This is for?

README.md
tancheng, line 1: Should we also mention we prefer the clang is also in llvm-19 version?

I am wondering, when we build LLVM-19.x from source, it is compiled using the built-in clang, which may not be llvm-19, right? Then if user install clang in llvm-19, should we update this https://github.com/coredac/dataflow/blob/ff914bc44d0c2cb5589ca962ff32cbe372ea9871/test/c2llvm2mlir/test.mlir#L5 to distinguish the system clang and the user-installed clang?

.github/workflows/main.yml
```diff
@@ -34,13 +34,16 @@ jobs:
         sudo apt-get install ccache
         sudo apt-get install lld
  
+    # install clang for LLVM build
+    - name: install clang
+      run: sudo apt-get install -y clang
```
tancheng, line 39: I may already ask, but what is this clang version?

README.md
yuqisun: Added this because Ninja cannot auto fix RPATH:

```
CMake Error at cmake/modules/AddLLVM.cmake:1029 (add_executable):
The install of the llvm-tblgen target requires changing an RPATH from the
build tree, but this is not supported with the Ninja generator unless on an
ELF-based or XCOFF-based platform. The CMAKE_BUILD_WITH_INSTALL_RPATH
variable may be set to avoid this relinking step.
Call Stack (most recent call first):
cmake/modules/TableGen.cmake:181 (add_llvm_executable)
utils/TableGen/CMakeLists.txt:32 (add_tablegen)
```

tancheng: Let's also have `DCMAKE_EXPORT_COMPILE_COMMANDS=ON` for `VS Code` users (to avoid warning in the IDE for include paths), I just set it up with copilot.

.github/workflows/main.yml
yuqisun: Default install version is 14.

README.md
yuqisun: Added.

README.md
yuqisun: How about now: Compile both clang and mlir and add note at the beginning of test.mlir to recommend user to add llvm-19 clang to PATH.

`-DLLVM_ENABLE_PROJECTS="clang;mlir"` --- To compile clang in llvm-19 but take more time.

README.md
```diff
@@ -54,3 +58,24 @@ Build LLVM & Neura
  $ /workspace/llvm-project/build/bin/llvm-lit *
 ```
 
+Docker-based Environment Setup
+--------------------------------------------------------
+**Option 1: Pull Pre-built Image**
+
+You can directly pull and use the pre-built Docker image:
+```sh
+$ docker pull cgra/neura:v1
```
yuqisun, line 67: Image is pushing in progress:

```
The push refers to repository [docker.io/cgra/neura]
3b5d14681ffd: Pushing [==================================================>]  12.92GB
346f14bf17b9: Layer already exists
```

PR #20, opened 2025-05-23
[feature] Handles terminator for insert mov

tancheng:
Just re-enable insert-mov in test, handling the terminator.

I plan to support ctrl-flow related mov in another PR, which might be non-trivial. @MeowMJ @HardyVon.


ShangkunLi (approved):
Looks good to me!

PR #23, opened 2025-05-26
Support ctrl-flow

tancheng:
- Include predicate operand for each op

TODO in the PR:
- [ ] Separate data_mov and ctrl_mov
- [ ] Remove `bb` but provide ctrl_mov with live-in analysis

### DIFF
--- a/include/NeuraDialect/NeuraOps.td
+++ b/include/NeuraDialect/NeuraOps.td
@@ -15,9 +15,9 @@ def Neura_ConstantOp : Op<NeuraDialect, "constant"> {
 def Neura_AddOp : Op<NeuraDialect, "add"> {
   let summary = "Integer addition operation";
   let opName = "add";
-  let arguments = (ins AnyInteger:$lhs, AnyInteger:$rhs);
+  let arguments = (ins AnyInteger:$lhs, AnyInteger:$rhs, Optional<AnyType>:$predicate);
   let results = (outs AnyInteger:$result);
-  // let assemblyFormat = "$lhs `,` $rhs attr-dict `:` type($result)";
+  // let assemblyFormat = "$lhs `,` $rhs `,` $predicate attr-dict `:` type($result)";
   let traits = [SameOperandsAndResultElementType];
 }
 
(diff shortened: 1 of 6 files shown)


test/neura/for_loop/test.mlir
```diff
@@ -1,5 +1,5 @@
 // Compiles the original kernel to mlir, then lower back to llvm, eventually binary.
-// RUN: clang++ -S -emit-llvm -O2 -o %t-kernel.ll kernel.cpp
+// RUN: clang++-19 -S -emit-llvm -O1 -o %t-kernel.ll kernel.cpp
```
tancheng, line 2: Hi @yuqisun, do you think this would work if I explicitly set clang to 19 (in our Github Action)?

yuqisun: I think `clang-19` should work (no `clang++-19`), and we need add clang in `-DLLVM_ENABLE_PROJECTS="clang;mlir"` and export PATH after build(3 hours), WDYT:

```
root@4c065acd966d:/cgra/dataflow/llvm-project/build/bin# ls -l
total 3883056
-rwxr-xr-x 1 root root   1053480 May 25 16:13 FileCheck
-rwxr-xr-x 1 root root     57304 May 25 16:13 UnicodeNameMappingGenerator
-rwxr-xr-x 1 root root    383424 May 25 18:50 amdgpu-arch
-rwxr-xr-x 1 root root       557 May 25 17:00 analyze-build
-rwxr-xr-x 1 root root    871376 May 25 18:48 apinotes-test
-rwxr-xr-x 1 root root  60994936 May 25 18:49 arcmt-test
-rwxr-xr-x 1 root root  60280968 May 25 18:50 bugpoint
-rwxr-xr-x 1 root root      9448 May 25 18:54 c-arcmt-test
-rwxr-xr-x 1 root root  64140184 May 25 18:54 c-index-test
lrwxrwxrwx 1 root root         8 May 25 18:50 clang -> clang-19
lrwxrwxrwx 1 root root         8 May 25 18:50 clang++ -> clang-19
-rwxr-xr-x 1 root root 172461440 May 25 18:50 clang-19
-rwxr-xr-x 1 root root 109895888 May 25 18:50 clang-check
lrwxrwxrwx 1 root root         8 May 25 18:50 clang-cl -> clang-19
lrwxrwxrwx 1 root root         8 May 25 18:50 clang-cpp -> clang-19
...
```

yuqisun (approved):

include/NeuraDialect/NeuraOps.td
```diff
@@ -140,12 +159,24 @@ def Neura_FMulFAddOp : Op<NeuraDialect, "fmul_fadd"> {
...
+    Op<NeuraDialect, mnemonic> {
+  let arguments = (ins AnyType:$input);
+  let results = (outs AnyType:$output);
+  // let assemblyFormat = "$input attr-dict `:` type($input) `->` type($output)";
+}
+
+def Neura_MovOp : NeuraBaseMov<"mov"> {
+  let summary = "Base move operation";
+}
+
```
ShangkunLi, line 215: What is the difference in functionalities between MovOp and DataMovOp

tancheng: `DataMov` is for data delivery while I also provide `CtrlMov`.

I initially plan to have both inherited from same base Op, however, failed due to some cmake or tablegen issue. And later I found I need CtrlMov to represent the backward flow. So in the latest commit in this PR, I just give two separate MovOp.

Issue #13, opened 2025-05-17
[P2] All ops from MLIR.LLVM should be covered


tancheng:
Fadd's conversion is supported: https://github.com/coredac/dataflow/blob/main/lib/Conversion/LlvmToNeura/LlvmToNeuraPatterns.td

All other ops should be covered in the same way.

PR #24, opened 2025-06-01
support llvm_fsub

Yfeng-44:
regarding #13 , I will add more Ops if this looks fine.

Please let me know if this is not what we are looking for.

Thanks!

### DIFF
--- a/lib/Conversion/LlvmToNeura/LlvmToNeuraPatterns.td
+++ b/lib/Conversion/LlvmToNeura/LlvmToNeuraPatterns.td
@@ -3,7 +3,7 @@ include "mlir/IR/PatternBase.td"
 include "mlir/Dialect/LLVMIR/LLVMOps.td"
 include "NeuraDialect/NeuraOps.td"
 
-// Floating point binary operations
+// Floating point binary operations.
 def : Pat<
   (LLVM_FAddOp $lhs, $rhs, $_fastmath),
   (Neura_FAddOp $lhs, $rhs)
--- a/lib/Conversion/LlvmToNeura/LlvmToNeuraPatterns.td
+++ b/lib/Conversion/LlvmToNeura/LlvmToNeuraPatterns.td
@@ -3,9 +3,15 @@ include "mlir/IR/PatternBase.td"
 include "mlir/Dialect/LLVMIR/LLVMOps.td"
 include "NeuraDialect/NeuraOps.td"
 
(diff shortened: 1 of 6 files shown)


lib/Conversion/LlvmToNeura/LlvmToNeuraPatterns.td
```diff
@@ -3,11 +3,17 @@ include "mlir/IR/PatternBase.td"
...
 def : Pat<
   (LLVM_FAddOp $lhs, $rhs, $_fastmath),
   (Neura_FAddOp $lhs, $rhs)
 >;
 
+def : Pat<
+  (LLVM_FSubOp $lhs, $rhs, $_fastmath),
+  (Neura_FSubOp $lhs, $rhs)
+>;
+
```
tancheng, line 16: Yes. But FYI, we are moving such pattern conversion in the .cc: lib/Conversion/LlvmToNeura/LlvmToNeuraPass.cpp as shown in this PR: https://github.com/coredac/dataflow/pull/23/files#

The reason is we are making these op support additional operand like `predicate`.

But it is okay for you to add the conversion here. I can fix them in my PR.

test/neura/llvm_sub.mlir
```diff
@@ -0,0 +1,10 @@
+// RUN: mlir-neura-opt --assign-accelerator --lower-llvm-to-neura --insert-mov %s | FileCheck %s
```
tancheng, line 1: It is even better to add the new op into the interpreter/emulator: https://github.com/coredac/dataflow/blob/main/tools/neura-interpreter/neura-interpreter.cpp

Corresponding test: https://github.com/coredac/dataflow/blob/86aab3e66f46d1003fddb8b9c2f045ac2d12cf88/test/neura/interpreter/lower_and_interpreter.mlir#L25

The interpreter issue is here: https://github.com/coredac/dataflow/issues/5, no need to make it complete for all ops, but it might be great to add the per-op support when you add new one.

Yfeng-44: Ops miss this part totally. let me get into it and get back.

Yfeng-44:
thank you for your patient comments. Once this pr get merged, I will work on supporting more llvm ops. 

Thanks.

tancheng (approved):

tancheng:
LGTM, let's merge it once the github action passes :-)

PR #29, opened 2025-06-12
Prototype ctrl to data transformation

tancheng:
Prototype ctrl to data flow transformation
 - Insert Phi at the beginning of each basic block
 - Connecting the predecessors with the newly inserted phi
 - Remove basic block

TODO:
 - [ ] Insert phi into the entry block
 - [ ] Enable backward ctrl flow connection towards phi
   - [ ] Plan to rename `ctrl_mov` to `backward_mov`

### DIFF
--- a/lib/NeuraDialect/Transforms/LeveragePredicatedValuePass.cpp
+++ b/lib/NeuraDialect/Transforms/LeveragePredicatedValuePass.cpp
@@ -18,28 +18,25 @@ struct applyPredicatedDataType : public RewritePattern {
       : RewritePattern(MatchAnyOpTypeTag(), /*benefit=*/1, context) {}
 
   LogicalResult matchAndRewrite(Operation *op, PatternRewriter &rewriter) const override {
-    llvm::errs() << "Processing op: " << *op << "\n";
 
     // Skips if not a Neura op or already using predicated values.
     if (op->getDialect()->getNamespace() != "neura") {
-        llvm::errs() << "Skipping non-Neura op\n";
         return failure();
     }
 
     if (llvm::any_of(op->getResultTypes(), 
         [](Type t) { return mlir::isa<mlir::neura::PredicatedValue>(t); })) {
-        llvm::errs() << "Skipping already predicated op\n";
         return failure();
     }
 
     // Converts result types to predicated form.
     SmallVector<Type> newResults;
     for (Type t : op->getResultTypes()) {
-        auto predicatedTy = mlir::neura::PredicatedValue::get(
+        auto predicated_type = mlir::neura::PredicatedValue::get(
             op->getContext(),
             t,
             rewriter.getI1Type());
-        newResults.push_back(predicatedTy);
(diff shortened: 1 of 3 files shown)


lib/NeuraDialect/Transforms/LeveragePredicatedValuePass.cpp
```diff
@@ -18,28 +18,25 @@ struct applyPredicatedDataType : public RewritePattern {
       : RewritePattern(MatchAnyOpTypeTag(), /*benefit=*/1, context) {}
```
ShangkunLi, line 18: Defined this pattern, but seems not using it.

lib/NeuraDialect/Transforms/LeveragePredicatedValuePass.cpp
```diff
@@ -77,7 +73,26 @@ struct LeveragePredicatedValuePass
...
+        // skips the entry (first) block of the function.
+        if (block == &block->getParent()->front()) {
+          return;
+        }
+        for (BlockArgument arg : block->getArguments()) {
+          Type origType = arg.getType();
+
+          // Avoid double-wrapping if already predicated
+          if (llvm::isa<neura::PredicatedValue>(origType))
+            continue;
```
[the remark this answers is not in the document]
tancheng: You mean why we wanna convert type to predicated type across all bb/operations, right?

I think this is fundamentally necessary for both loop and if/else (loop is also type of ctrl flow, I am preparing the test and will also insert phi and predicated type for it in next PR). Both of them needs somehow notify consumer active or inactive (or even stop), including your high level affine loop's direct lowering, which requires a new data type and this predicated type is used to serve that.

From mlir.llvm perspective, everything is basic block (no difference between if/else and for loop), I would have to `replace data type` then `insert phi op` then `remove all bb`. `replace data type` needs to be done for all bb and ops, as the values are propagated across all producer-consumer chains, we cannot just replace a few of them; `insert phi op` and `remove all bb` need to be done in a single pass.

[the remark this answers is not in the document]
ShangkunLi: Got it! Thx!

lib/NeuraDialect/Transforms/LeveragePredicatedValuePass.cpp
tancheng: Removed in latest commit. Thanks! I probably encountered some issues when use/apply this type of pattern match, maybe related to the ordering. So I explicitly `walk()` ops in the topological order instead.

tancheng:
@ShangkunLi plz help approve if all your concerns are resolved :-)

lib/NeuraDialect/Transforms/LeveragePredicatedValuePass.cpp
tancheng: You can click the `approve` button then:

Screenshot 2025-06-12 at 9 24 43 AM

PR #37, opened 2025-06-15
add support of neura.PhiOp in interpreter

Yfeng-44:
Add neura.PhiOp interpreter support

to verify this patch `./build/tools/neura-interpreter/neura-interpreter /home/yuan/neura_dataflow/dataflow/test/neura/interpreter/phiOp_interpret.mlir`

### DIFF
--- a/test/neura/interpreter/phiOp_interpret.mlir
+++ b/test/neura/interpreter/phiOp_interpret.mlir
@@ -1,0 +1,20 @@
+// RUN: neura-interpreter %s | FileCheck %s
+
+module {
+  func.func @test_phi() -> f32 {
+    %true_path_val = "neura.constant"() <{
+      value = 10.0 : f32,
+      predicate = true
+    }> : () -> f32
+
+    %false_path_val = "neura.constant"() <{
+      value = 20.0 : f32,
+      predicate = false
+    }> : () -> f32
+
+    %result = "neura.phi"(%true_path_val, %false_path_val) : (f32, f32) -> f32
+
+    return %result : f32
+    // CHECK: [neura-interpreter] Output: 10.000000
+  }
+}
--- a/tools/neura-interpreter/neura-interpreter.cpp
+++ b/tools/neura-interpreter/neura-interpreter.cpp
@@ -95,6 +95,17 @@ int main(int argc, char **argv) {
         
         valueMap[constOp.getResult()] = val;
 
+      } else if (auto phiOp = dyn_cast<neura::PhiOp>(op)) {
+        PredicatedData result{0.0f, false}; // Default to a false predicate.
+        // Find the one operand with a true predicate.
+        for (Value operand : phiOp.getOperands()) {
+            auto incoming = valueMap[operand];
+            if (incoming.predicate) {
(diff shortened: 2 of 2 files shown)


Yfeng-44:
#28

tools/neura-interpreter/neura-interpreter.cpp
```diff
@@ -95,6 +95,17 @@ int main(int argc, char **argv) {
...
         valueMap[constOp.getResult()] = val;
 
+      } else if (auto phiOp = dyn_cast<neura::PhiOp>(op)) {
+        PredicatedData result{0.0f, false}; // Default to a false predicate.
+        // Find the one operand with a true predicate.
+        for (Value operand : phiOp.getOperands()) {
+            auto incoming = valueMap[operand];
+            if (incoming.predicate) {
+                result = incoming;
+                break; // Found the active value.
```
tancheng, line 105: Move the comment above the `if`.

test/neura/interpreter/phiOp_interpret.mlir
```diff
@@ -0,0 +1,20 @@
+// RUN: neura-interpreter %s | FileCheck %s
+
```
tancheng, line 2: Can we run mlir-neura-opt with `--assign-accelerator --leverage-predicated-value` so the IR would contain `PredicatedValue` as https://github.com/coredac/dataflow/blob/64e7b3537cc5de007ec81b706c67cb20f0ce2860/test/neura/ctrl/branch_no_arg.mlir#L62
and then run `neura-interpreter` again to double check?

tools/neura-interpreter/neura-interpreter.cpp
```diff
@@ -95,6 +95,17 @@ int main(int argc, char **argv) {
         
         valueMap[constOp.getResult()] = val;
 
+      } else if (auto phiOp = dyn_cast<neura::PhiOp>(op)) {
+        PredicatedData result{0.0f, false}; // Default to a false predicate.
+        // Find the one operand with a true predicate.
+        for (Value operand : phiOp.getOperands()) {
```
tancheng, line 101: Shouldn't we `assert` there must be **at most one** operand with predicate as `true`?

test/neura/interpreter/phiOp_interpret.mlir
tancheng, line 1: Rename the test as `phi_interpret.mlir`.

tools/neura-interpreter/neura-interpreter.cpp
tancheng: Plz hold on, let me re-think the design. Its current form cannot support the diagram shown in README: https://github.com/tancheng/CGRA-Mapper

tancheng: I introduced `grant_predicate` and `grant_once` in https://github.com/coredac/dataflow/pull/39, so we don't need to complicate `phi`'s definition for now, so you are good to go: "assert there must be at most one operand with predicate as true".

Later on, we can introduce new op to fuse the (`phi` and `grant_once` and `const`) if needed.

PR #47, opened 2025-06-18
Add neura-compiler

ShangkunLi:
Now you can define some pass pipelines to run the code end-to-end.

### DIFF
--- a/tools/neura-compiler/neura-compiler.cpp
+++ b/tools/neura-compiler/neura-compiler.cpp
@@ -1,0 +1,32 @@
+// neura-compiler.cpp
+
+#include "mlir/Dialect/Affine/IR/AffineOps.h"
+#include "mlir/Dialect/DLTI/DLTI.h"
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+#include "mlir/InitAllDialects.h"
+#include "mlir/InitAllPasses.h"
+#include "mlir/Support/FileUtilities.h"
+#include "mlir/Support/LogicalResult.h"
+#include "mlir/Tools/mlir-opt/MlirOptMain.h"
+
+#include "Conversion/ConversionPasses.h"
+#include "NeuraDialect/NeuraDialect.h"
+#include "NeuraDialect/NeuraPasses.h"
+
+int main(int argc, char **argv) {
+  // Registers MLIR dialects.
+  mlir::DialectRegistry registry;
+  registry.insert<mlir::neura::NeuraDialect>();
+  registry.insert<mlir::func::FuncDialect>();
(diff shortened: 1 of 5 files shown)


tancheng:
Thanks @ShangkunLi, can we enable this feature on [arith ir](https://github.com/coredac/dataflow/blob/main/test/arith2neura/add.mlir) and [mlir.llvm ir](https://github.com/coredac/dataflow/blob/main/test/neura/ctrl/branch.mlir) to verify the generated dataflow IR?

test/arith2neura/add.mlir
```diff
@@ -1,8 +1,11 @@
-// RUN: mlir-neura-opt --lower-arith-to-neura %s | FileCheck %s
+// RUN: neura-compiler --neura-conversion %s | FileCheck %s --check-prefix=COMPILER
+// RUN: mlir-neura-opt --lower-arith-to-neura %s | FileCheck %s --check-prefix=OPT
```
tancheng, line 2: Excellent! Can we have the same thing for https://github.com/coredac/dataflow/blob/main/test/neura/ctrl/branch.mlir? Or any blocker for that?

tools/neura-compiler/neura-compiler.cpp
```diff
@@ -0,0 +1,32 @@
...
+  mlir::DialectRegistry registry;
+  registry.insert<mlir::neura::NeuraDialect>();
+  registry.insert<mlir::func::FuncDialect>();
+  registry.insert<mlir::arith::ArithDialect>();
+  registry.insert<mlir::DLTIDialect>();
+  registry.insert<mlir::LLVM::LLVMDialect>();
+  registry.insert<mlir::affine::AffineDialect>();
+  registry.insert<mlir::memref::MemRefDialect>();
+
+  mlir::neura::registerNeuraConversionPassPipeline();
```
tancheng, line 27: Excellent! I am wondering how could we use flags the similar way as the `neura-opt`. Specifically, for a compiler, it is popular/common to enable/disable some passes (by default pass A is disabled, a flag `--enable-A` can enable it; or by default pass B is enabled, a flag `--disable-B` can disable it). For our compiler, all these passes should be registered here anyways I guess, but how to make each of them be able to digest specific flag related to the corresponding pass? Plz file an issue for this request, if you think this makes sense (i think it would be necessary especially for different patterns fusion -- making them under user's control).

tancheng (approved):

tools/neura-compiler/neura-compiler.cpp
ShangkunLi: Filed an issue #48 

May need more discussion.

PR #50, opened 2025-06-19
Re-organize mapping util functions

tancheng:
- Provide resMII calculation
 - Re-org util funcs
 - Extreme simple arch spec structure is provided

TODO:
 - [ ] We need model the dataflow accelerator arch during mapping based on a protobuff/json that can be used by both RTL and compiler: https://github.com/coredac/dataflow/issues/52

### DIFF
--- a/lib/NeuraDialect/Transforms/MapToAcceleratorPass.cpp
+++ b/lib/NeuraDialect/Transforms/MapToAcceleratorPass.cpp
@@ -4,93 +4,19 @@
 #include "NeuraDialect/NeuraOps.h"
 #include "NeuraDialect/NeuraTypes.h"
 #include "NeuraDialect/NeuraPasses.h"
+#include "NeuraDialect/mapping/mapping_util.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/Pass/Pass.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 
 using namespace mlir;
+using namespace mlir::neura;
 
 #define GEN_PASS_DEF_MapToAccelerator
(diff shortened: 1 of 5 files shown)


lib/NeuraDialect/Transforms/MapToAcceleratorPass.cpp
```diff
@@ -127,10 +53,17 @@ struct MapToAcceleratorPass
                     << longest->length << "):\n";
         for (Operation *op : longest->operations)
           op->print(llvm::errs()), llvm::errs() << "\n";
-        IntegerAttr mii_attr = IntegerAttr::get(
+        IntegerAttr rec_mii_attr = IntegerAttr::get(
             IntegerType::get(func.getContext(), 32), longest->length);
-        func->setAttr("RecMII", mii_attr);
+        func->setAttr("RecMII", rec_mii_attr);
       }
```
ShangkunLi, line 59: For a function that contains many loops (like the Bert model provided in samples), should we create a single DFG for all the DFGs?

Or we can utilize loop fusion to fuse some loops and create a DFG for each single loop?

tancheng: Ah, I thought func-level is easier, similar to our previous MLIR-CGRA project, we just need to get the live-in, embed them as func_args, and move all operations into the func op.

For the prologs, they can be moved out of the func, and passed in as argument, so we don't need to map them on the accelerator.

Anyways, we need a region or func to indicate the scope that the accelerator targets, from the firmware and runtime perspective. Func op sounds suitable for this purpose. We can discuss this during our 1:1.

ShangkunLi (approved):

tancheng:
I will create an `Architecture` folder to formally construct the CGRA topology for mapping ASAP, in the next PR. Thanks!

PR #56, opened 2025-06-24
Construct MRRG

tancheng:
Prototyping MRRG construction.


lib/NeuraDialect/CMakeLists.txt
```diff
@@ -19,6 +19,9 @@ add_mlir_dialect_library(MLIRNeura
         Neura.cpp
         NeuraTypes.cpp
         NeuraPasses.cpp
+        Mapping/mapping_util.cpp
+        Mapping/MappingState.cpp
+        Architecture/Architecture.cpp
```
ShangkunLi, line 24: Maybe we can create a library for those files in `Mapping` instead of adding them one by one.

tancheng: Need your help on this :-)

Issue #59, opened 2025-06-28
[P1] Investigate/refactor the cost/award estimation for mapping


tancheng:
https://github.com/coredac/dataflow/pull/58 enables the simple cost/award estimation for MRRG mapping.

However, it is now super rough and when I try to include the utilization (e.g., lower utilized tile should be prioritized for placement of an operation), the compiled II unfortunately becomes worse.

For now, the simple for loop's `RecMII = 4`, `ResMII = 1`, but `CompiledII = 6`: https://github.com/coredac/dataflow/blob/3a3c5546a7b6c0cf2bbef7692474a8570a2b6922/test/neura/ctrl/branch_for.mlir#L141
Our goal is to decrease `CompiledII` from 6 to 4. Note that the pattern matching is orthogonal, which can make `RecMII = 1` hopefully.

TODO:
 - [ ] Figure out why more consideration leads to worse compiled II.
 - [ ] Include the utilization of HW resources into the heuristic mapping.
 - [ ] Enable backtrack for operation placement, i.e., if mapping is not valid when an operation is being placed or when a move operation is being routed, we need to backtrack to the previous/producer operation's placement.

ShangkunLi:
After adding the mixed heuristic of ALAP & ASAP, for now, the bottleneck seems to come from choosing the tile for p&r.

For example, in the following mapping result, there is an unbalanced PE utilization, causing routing congestion, thus resulting in an increased II (`CompiledII=4, RecMII=1, ResMII=2`)

Specifically,
1. The `tile (id = 6, x = 2, y = 1)` is used for 4 operations:
```
%nextindex, %valid = neura.loop_control(parent_valid = %8, start = %9, end = %10, step = %11) {iterationType = "increment", mapping_locs = [{id = 6 : i32, resource = "tile", time_step = 3 : i32, x = 2 : i32, y = 1 : i32}]} : !neura.data<i1, i1>, !neura.data<i64, i1>, !neura.data<i64, i1>, !neura.data<i64, i1> -> !neura.data<i64, i1>, !neura.data<i1, i1>
%13 = "neura.not"(%12) {mapping_locs = [{id = 6 : i32, resource = "tile", time_step = 6 : i32, x = 2 : i32, y = 1 : i32}]} : (!neura.data<i1, i1>) -> !neura.data<i1, i1>
[...]

PR #60, opened 2025-06-29
Refactor the Ctrl to Data Flow Implementation Logic

ShangkunLi:
In this PR, an edge-based control flow to data flow transform pass is implemented.

Basically, we can categorize all the edges in CFG into the following 8 cases:
1. Backward cond_br edges with values.
2. Backward br edges with values.
3. Backward cond_br without values.
4. Backward br without value.
5. Forward cond_br edges with values.
6. Forward br edges with values.
7. Forward cond_br edges without values.
8. Forward br edges without values.

Cases 3 and 4 do not appear in the current benchmarks. Since they correspond to control flow jumps for statements like `goto`, we do not consider these cases for now.

The transform is implemented based on the remaining six edges.

For the target block of the edge in case 7, we chose to `grant_predicate` all results in this block based on condition to ensure correctness. For example:

target block `bb2`:
```
^bb1(%4: !neura.data<i64, i1>):  // 2 preds: ^bb0, ^bb5
    %5 = "neura.cast"(%4) <{cast_type = "int_to_index"}> : (!neura.data<i64, i1>) -> !neura.data<index, i1>
    %6 = "neura.icmp"(%5, %1) <{cmpType = "slt"}> : (!neura.data<index, i1>, !neura.data<index, i1>) -> !neura.data<i1, i1>
    neura.cond_br %6 : !neura.data<i1, i1> then to ^bb2 else to ^bb6
^bb2:  // pred: ^bb1
    %7 = "neura.cast"(%2) <{cast_type = "index_to_int"}> : (!neura.data<index, i1>) -> !neura.data<i64, i1>
    neura.br %7 : !neura.data<i64, i1> to ^bb3
```
transformed ir:
```
%11 = "neura.icmp"(%10, %3) <{cmpType = "slt"}> : (!neura.data<index, i1>, !neura.data<index, i1>) -> !neura.data<i1, i1>
%12 = "neura.not"(%11) : (!neura.data<i1, i1>) -> !neura.data<i1, i1>
%13 = "neura.cast"(%5) <{cast_type = "index_to_int"}> : (!neura.data<index, i1>) -> !neura.data<i64, i1>
%14 = neura.grant_predicate %13, %11 : !neura.data<i64, i1>, !neura.data<i1, i1> -> !neura.data<i64, i1>
```

We `grant_predicate` the result of `bb2` -- `%13` with the condition of its pred block.

### DIFF
--- a/lib/NeuraDialect/Transforms/TransformCtrlToDataFlowPass.cpp
+++ b/lib/NeuraDialect/Transforms/TransformCtrlToDataFlowPass.cpp
@@ -509,7 +509,7 @@ void createReserveAndPhiOps(func::FuncOp &func, ControlFlowInfo &ctrl_info,
         }
       }
 
-      // Apply grant_predicate for each live-in value.
+      // Applies grant_predicate for each live-in value.
       for (Value live_in_value : live_in_values) {
         // Finds the earliest use of the live-in value.
         Operation *earliest_use = nullptr;
--- a/lib/NeuraDialect/Transforms/TransformCtrlToDataFlowPass.cpp
+++ b/lib/NeuraDialect/Transforms/TransformCtrlToDataFlowPass.cpp
@@ -264,14 +264,14 @@ void createReserveAndPhiOps(func::FuncOp &func, ControlFlowInfo &ctrl_info,
   DenseMap<BlockArgument, SmallVector<Value>> arg_to_phi_operands;
 
   for (auto &edge : ctrl_info.all_edges) {
-    Block *source = edge->source;
     Block *target = edge->target;
 
     // Type 1 & 2: Backward cond_br/br edges with values.
     if (edge->is_back_edge && !edge->passed_values.empty()) {
       if (edge->passed_values.size() != target->getNumArguments()) {
-        llvm::errs() << "[ctrl2data] Error: Number of passed values does not match "
-                        "target block arguments.\n";
+        llvm::errs()
+            << "[ctrl2data] Error: Number of passed values does not match "
+               "target block arguments.\n";
         assert(false);
       }
       for (BlockArgument arg : target->getArguments()) {
@@ -279,10 +279,12 @@ void createReserveAndPhiOps(func::FuncOp &func, ControlFlowInfo &ctrl_info,
       }
     }
     // Type 3 & 4: Forward cond_br/br edges with values.
-    else if (!edge->is_back_edge && !edge->passed_values.empty()) {;
+    else if (!edge->is_back_edge && !edge->passed_values.empty()) {
+      ;
       if (edge->passed_values.size() != target->getNumArguments()) {
-        llvm::errs() << "[ctrl2data] Error: Number of passed values does not match "
-                        "target block arguments.\n";
+        llvm::errs()
+            << "[ctrl2data] Error: Number of passed values does not match "
+               "target block arguments.\n";
         assert(false);
       }
       for (BlockArgument arg : target->getArguments()) {
@@ -373,15 +375,14 @@ void createReserveAndPhiOps(func::FuncOp &func, ControlFlowInfo &ctrl_info,
       // No need to create a phi operation if there's only one operand.
 
       if (phi_operands.size() == 1) {
-        llvm::errs() << phi_operands[0] << "\n";
         arg_to_phi_result[arg] = phi_operands[0];
         arg.replaceAllUsesWith(phi_operands[0]);
       }
       continue;
     }
 
-
-    // Handles the blcockargument with/without reserve seperately (different insertion points).
+    // Handles the blcockargument with/without reserve seperately (different
+    // insertion points).
     if (arg_to_reserve.count(arg)) {
       Value reserve_value = arg_to_reserve[arg];
       builder.setInsertionPointAfter(reserve_value.getDefiningOp());
@@ -403,6 +404,7 @@ void createReserveAndPhiOps(func::FuncOp &func, ControlFlowInfo &ctrl_info,
     }
   }
 
+  llvm::errs() << func << "\n";
   // ================================================
   // Step 5: Handles Forward cond_br edges without values.
   // ================================================
@@ -423,34 +425,61 @@ void createReserveAndPhiOps(func::FuncOp &func, ControlFlowInfo &ctrl_info,
       conditions.push_back(condition);
     }
 
-
     // Unsupported case: multiple conditions for a single block.
     // TODO: Adds support if needed.
     if (conditions.size() > 1) {
-      llvm::errs()
-          << "[ctrl2data] Unsupported case: multiple conditions for a single block: "
-          << *target << "\n";
+      llvm::errs() << "[ctrl2data] Unsupported case: multiple conditions for a "
+                      "single block: "
+                   << *target << "\n";
       assert(false);
     }
 
     if (target->getArguments().empty()) {
+      // Grants predicate for all the live-in values in the target block.
+      DenseSet<Value> live_in_values;
       for (Operation &op : target->getOperations()) {
-        if (op.hasTrait<OpTrait::IsTerminator>() ||
-            isa<neura::PhiOp, neura::ReserveOp, neura::CtrlMovOp,
-                neura::GrantPredicateOp, neura::GrantAlwaysOp,
-                neura::GrantOnceOp, neura::NotOp>(op)) {
-          continue;
+        for (Value operand : op.getOperands()) {
+          if (operand.getDefiningOp() &&
+              operand.getDefiningOp()->getBlock() != target &&
+              !isa<neura::ReserveOp>(operand.getDefiningOp())) {
+            live_in_values.insert(operand);
+          }
+        }
+      }
+
+      // Apply grant_predicate for each live-in value.
+      for (Value live_in_value : live_in_values) {
+        // Finds the earliest use of the live-in value.
+        Operation *earliest_use = nullptr;
+        for (Operation &op : target->getOperations()) {
+          for (Value operand : op.getOperands()) {
+            if (operand == live_in_value) {
+              earliest_use = &op;
+              break;
+            }
+          }
+          if (earliest_use) {
+            break;
+          }
         }
 
-        builder.setInsertionPointAfter(&op);
+        if (earliest_use) {
+          builder.setInsertionPoint(earliest_use);
+        } else {
+          builder.setInsertionPointToStart(target);
+        }
 
-        for (Value result : op.getResults()) {
-          if (isa<neura::PredicatedValue>(result.getType())) {
-            // Creates a predicated value for the operation result.
-            Value predicated_value = builder.create<neura::GrantPredicateOp>(
-                op.getLoc(), result.getType(), result, conditions.front());
-            result.replaceAllUsesExcept(predicated_value,
-                                        predicated_value.getDefiningOp());
+        // Creates predicated version of the live-in value
+        Value predicated_value = builder.create<neura::GrantPredicateOp>(
+            live_in_value.getLoc(), live_in_value.getType(), live_in_value,
+            conditions[0]);
+
+        // Replace uses of the live-in value within this block only.
+        for (OpOperand &use :
+             llvm::make_early_inc_range(live_in_value.getUses())) {
+          if (use.getOwner()->getBlock() == target &&
+              use.getOwner() != predicated_value.getDefiningOp()) {
+            use.set(predicated_value);
           }
         }
       }
@@ -494,7 +523,8 @@ void transformControlFlowToDataFlow(func::FuncOp &func,
     }
   }
 
-  // Moves all operations from blocks to the entry block before the terminator.
+  // Moves all operations from blocks to the entry block before the
+  // terminator.
   for (Block *block : blocks_to_flatten) {
     auto &ops = block->getOperations();
     while (!ops.empty()) {
--- a/lib/NeuraDialect/Transforms/TransformCtrlToDataFlowPass.cpp
+++ b/lib/NeuraDialect/Transforms/TransformCtrlToDataFlowPass.cpp
@@ -1,11 +1,19 @@
 #include "Common/AcceleratorAttrs.h"
 #include "NeuraDialect/NeuraDialect.h"
 #include "NeuraDialect/NeuraOps.h"
-#include "NeuraDialect/NeuraTypes.h"
 #include "NeuraDialect/NeuraPasses.h"
-#include "mlir/IR/PatternMatch.h"
-#include "mlir/Pass/Pass.h"
+#include "NeuraDialect/NeuraTypes.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/IR/Block.h"
+#include "mlir/IR/Dominance.h"
+#include "mlir/IR/Location.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/IR/Value.h"
+#include "mlir/Pass/Pass.h"
+#include "llvm/ADT/STLExtras.h"
+#include "llvm/Support/Casting.h"
+#include "llvm/Support/Error.h"
+#include "llvm/Support/raw_ostream.h"
 
 using namespace mlir;
 
@@ -32,7 +40,7 @@ void GrantPredicateInEntryBlock(Block *entry_block, OpBuilder &builder) {
 
         // Case 1: Operand of a branch/cond_br → grant_once
         if (isa<neura::Br, neura::CondBr>(user)) {
-            used_in_branch = true;
+          used_in_branch = true;
         }
 
         // Case 2: Used directly in other blocks → grant_always
@@ -56,7 +64,8 @@ void GrantPredicateInEntryBlock(Block *entry_block, OpBuilder &builder) {
       continue;
 
     builder.setInsertionPointAfter(def_op);
-    auto granted = builder.create<neura::GrantOnceOp>(def_op->getLoc(), val.getType(), val);
+    auto granted = builder.create<neura::GrantOnceOp>(def_op->getLoc(),
+                                                      val.getType(), val);
 
     // Replaces uses in branch ops.
     for (OpOperand &use : llvm::make_early_inc_range(val.getUses())) {
@@ -74,7 +83,8 @@ void GrantPredicateInEntryBlock(Block *entry_block, OpBuilder &builder) {
       continue;
 
     builder.setInsertionPointAfter(def_op);
-    auto granted = builder.create<neura::GrantAlwaysOp>(def_op->getLoc(), val.getType(), val);
+    auto granted = builder.create<neura::GrantAlwaysOp>(def_op->getLoc(),
+                                                        val.getType(), val);
 
     // Replaces direct external uses (not in entry block, not in branch ops).
     for (OpOperand &use : llvm::make_early_inc_range(val.getUses())) {
@@ -88,8 +98,9 @@ void GrantPredicateInEntryBlock(Block *entry_block, OpBuilder &builder) {
 }
 
 // Returns blocks in post-order traversal order.
-void getBlocksInPostOrder(Block *startBlock, SmallVectorImpl<Block *> &postOrder,
-                         DenseSet<Block *> &visited) {
+void getBlocksInPostOrder(Block *startBlock,
+                          SmallVectorImpl<Block *> &postOrder,
+                          DenseSet<Block *> &visited) {
   if (!visited.insert(startBlock).second)
     return;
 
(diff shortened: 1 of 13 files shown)


tancheng:
Excellent summarization @ShangkunLi.

> For the target block of the edge in case 7, we chose to grant_predicate all results in this block based on condition to ensure correctness.

Shouldn't we grant_predicate **all live-out** results of `bb1`? Your example shows the `bb2`'s live-out is predicated based on `bb1`'s condition. This is some kind of indirect? WDYT?

tancheng:
Ah I got it. `bb2` is dominated by `bb1` but the its `live-in` is from another block. In such case, shouldn't we grant_predicate to `bb2`'s live-in?

tancheng:
> > Ah I got it. `bb2` is dominated by `bb1` but the its `live-in` is from another block. In such case, shouldn't we grant_predicate to `bb2`'s live-in?
> 
> You mean grant_predicate all the live-in values used in the block? But in many benchmarks, most live-ins are granted always. Like `%2` in `bb2`, `%2` and `%0` in `bb4`. So do we need to grant_predicate such grant_always values? Or just grant_predicat each result of operations in such block?
> 
> ```
> func.func @_Z10bert_node1PA1_A1_A1_A1_A128_bPA1_A128_S1_(%arg0: memref<?x1x1x1x1x128xi8>, %arg1: memref<?x1x128x1x1x128xi8>) attributes {accelerator = "neura", llvm.linkage = #llvm.linkage<external>} {
>     %0 = "neura.constant"() <{value = 1 : index}> : () -> index
>     %1 = "neura.constant"() <{value = 128 : index}> : () -> index
>     %2 = "neura.constant"() <{value = 0 : index}> : () -> index
>     %3 = "neura.cast"(%2) <{cast_type = "index_to_int"}> : (index) -> i64
>     neura.br %3 : i64 to ^bb1
> ^bb1(%4: i64):  // 2 preds: ^bb0, ^bb5
>     %5 = "neura.cast"(%4) <{cast_type = "int_to_index"}> : (i64) -> index
>     %6 = "neura.icmp"(%5, %1) <{cmpType = "slt"}> : (index, index) -> i1
>     neura.cond_br %6 : i1 then to ^bb2 else to ^bb6
> ^bb2:  // pred: ^bb1
>     %7 = "neura.cast"(%2) <{cast_type = "index_to_int"}> : (index) -> i64
>     neura.br %7 : i64 to ^bb3
> ^bb3(%8: i64):  // 2 preds: ^bb2, ^bb4
>     %9 = "neura.cast"(%8) <{cast_type = "int_to_index"}> : (i64) -> index
>     %10 = "neura.icmp"(%9, %1) <{cmpType = "slt"}> : (index, index) -> i1
>     neura.cond_br %10 : i1 then to ^bb4 else to ^bb5
> ^bb4:  // pred: ^bb3
>     %11 = neura.load_indexed %arg0[%2, %2, %2, %2, %2, %9 : index, index, index, index, index, index] memref<?x1x1x1x1x128xi8> : i8
>     neura.store_indexed %11 to %arg1[%2, %2, %5, %2, %2, %9 : index, index, index, index, index, index] memref<?x1x128x1x1x128xi8> : i8
>     %12 = "neura.add"(%9, %0) : (index, index) -> index
>     %13 = "neura.cast"(%12) <{cast_type = "index_to_int"}> : (index) -> i64
>     neura.br %13 : i64 to ^bb3
> ^bb5:  // pred: ^bb3
>     %14 = "neura.add"(%5, %0) : (index, index) -> index
>     %15 = "neura.cast"(%14) <{cast_type = "index_to_int"}> : (index) -> i64
>     neura.br %15 : i64 to ^bb1
> ^bb6:  // pred: ^bb1
>     "neura.return"() : () -> ()
> }
> ```

I believe we can `grant_predicate` further on the already `grant_always`, i.e., all the live-in of `bb2`, rather than the result of `bb2`.
 - I think the rule is to (use `bb1` with `cond_br` as an example) `grant_predicate` all `bb1`'s live-out and all true-successor bb's live-in, WDYT?
 - If one `const`/`grant_always` is used by multiple BBs being dominated by different `conditions`, we can `grant_predicate` them based on the specific `condition`
 - For now, we blindly give `grant_always` for all the const inside the entry block, we later can fuse the `grant_always` -> `grant_predicate`

How does this sound?

ShangkunLi:
> > > Ah I got it. `bb2` is dominated by `bb1` but the its `live-in` is from another block. In such case, shouldn't we grant_predicate to `bb2`'s live-in?
> > 
> > 
> > You mean grant_predicate all the live-in values used in the block? But in many benchmarks, most live-ins are granted always. Like `%2` in `bb2`, `%2` and `%0` in `bb4`. So do we need to grant_predicate such grant_always values? Or just grant_predicat each result of operations in such block?
> > ```
> > func.func @_Z10bert_node1PA1_A1_A1_A1_A128_bPA1_A128_S1_(%arg0: memref<?x1x1x1x1x128xi8>, %arg1: memref<?x1x128x1x1x128xi8>) attributes {accelerator = "neura", llvm.linkage = #llvm.linkage<external>} {
> >     %0 = "neura.constant"() <{value = 1 : index}> : () -> index
> >     %1 = "neura.constant"() <{value = 128 : index}> : () -> index
> >     %2 = "neura.constant"() <{value = 0 : index}> : () -> index
> >     %3 = "neura.cast"(%2) <{cast_type = "index_to_int"}> : (index) -> i64
> >     neura.br %3 : i64 to ^bb1
> > ^bb1(%4: i64):  // 2 preds: ^bb0, ^bb5
> >     %5 = "neura.cast"(%4) <{cast_type = "int_to_index"}> : (i64) -> index
> >     %6 = "neura.icmp"(%5, %1) <{cmpType = "slt"}> : (index, index) -> i1
> >     neura.cond_br %6 : i1 then to ^bb2 else to ^bb6
> > ^bb2:  // pred: ^bb1
> >     %7 = "neura.cast"(%2) <{cast_type = "index_to_int"}> : (index) -> i64
> >     neura.br %7 : i64 to ^bb3
> > ^bb3(%8: i64):  // 2 preds: ^bb2, ^bb4
> >     %9 = "neura.cast"(%8) <{cast_type = "int_to_index"}> : (i64) -> index
> >     %10 = "neura.icmp"(%9, %1) <{cmpType = "slt"}> : (index, index) -> i1
> >     neura.cond_br %10 : i1 then to ^bb4 else to ^bb5
> > ^bb4:  // pred: ^bb3
> >     %11 = neura.load_indexed %arg0[%2, %2, %2, %2, %2, %9 : index, index, index, index, index, index] memref<?x1x1x1x1x128xi8> : i8
> >     neura.store_indexed %11 to %arg1[%2, %2, %5, %2, %2, %9 : index, index, index, index, index, index] memref<?x1x128x1x1x128xi8> : i8
> >     %12 = "neura.add"(%9, %0) : (index, index) -> index
> >     %13 = "neura.cast"(%12) <{cast_type = "index_to_int"}> : (index) -> i64
> >     neura.br %13 : i64 to ^bb3
> > ^bb5:  // pred: ^bb3
> >     %14 = "neura.add"(%5, %0) : (index, index) -> index
> >     %15 = "neura.cast"(%14) <{cast_type = "index_to_int"}> : (index) -> i64
> >     neura.br %15 : i64 to ^bb1
> > ^bb6:  // pred: ^bb1
> >     "neura.return"() : () -> ()
> > }
> > ```
> 
> I believe we can `grant_predicate` further on the already `grant_always`, i.e., all the live-in of `bb2`, rather than the result of `bb2`.
> 
> * I think the rule is to (use `bb1` with `cond_br` as an example) `grant_predicate` all `bb1`'s live-out and all true-successor bb's live-in, WDYT?
> * If one `const`/`grant_always` is used by multiple BBs being dominated by different `conditions`, we can `grant_predicate` them based on the specific `condition`
> * For now, we blindly give `grant_always` for all the const inside the entry block, we later can fuse the `grant_always` -> `grant_predicate`
> 
> How does this sound?

Sure, this sounds more robust!

ShangkunLi:
> > > Ah I got it. `bb2` is dominated by `bb1` but the its `live-in` is from another block. In such case, shouldn't we grant_predicate to `bb2`'s live-in?
> > 
> > 
> > You mean grant_predicate all the live-in values used in the block? But in many benchmarks, most live-ins are granted always. Like `%2` in `bb2`, `%2` and `%0` in `bb4`. So do we need to grant_predicate such grant_always values? Or just grant_predicat each result of operations in such block?
> > ```
> > func.func @_Z10bert_node1PA1_A1_A1_A1_A128_bPA1_A128_S1_(%arg0: memref<?x1x1x1x1x128xi8>, %arg1: memref<?x1x128x1x1x128xi8>) attributes {accelerator = "neura", llvm.linkage = #llvm.linkage<external>} {
> >     %0 = "neura.constant"() <{value = 1 : index}> : () -> index
> >     %1 = "neura.constant"() <{value = 128 : index}> : () -> index
> >     %2 = "neura.constant"() <{value = 0 : index}> : () -> index
> >     %3 = "neura.cast"(%2) <{cast_type = "index_to_int"}> : (index) -> i64
> >     neura.br %3 : i64 to ^bb1
> > ^bb1(%4: i64):  // 2 preds: ^bb0, ^bb5
> >     %5 = "neura.cast"(%4) <{cast_type = "int_to_index"}> : (i64) -> index
> >     %6 = "neura.icmp"(%5, %1) <{cmpType = "slt"}> : (index, index) -> i1
> >     neura.cond_br %6 : i1 then to ^bb2 else to ^bb6
> > ^bb2:  // pred: ^bb1
> >     %7 = "neura.cast"(%2) <{cast_type = "index_to_int"}> : (index) -> i64
> >     neura.br %7 : i64 to ^bb3
> > ^bb3(%8: i64):  // 2 preds: ^bb2, ^bb4
> >     %9 = "neura.cast"(%8) <{cast_type = "int_to_index"}> : (i64) -> index
> >     %10 = "neura.icmp"(%9, %1) <{cmpType = "slt"}> : (index, index) -> i1
> >     neura.cond_br %10 : i1 then to ^bb4 else to ^bb5
> > ^bb4:  // pred: ^bb3
> >     %11 = neura.load_indexed %arg0[%2, %2, %2, %2, %2, %9 : index, index, index, index, index, index] memref<?x1x1x1x1x128xi8> : i8
> >     neura.store_indexed %11 to %arg1[%2, %2, %5, %2, %2, %9 : index, index, index, index, index, index] memref<?x1x128x1x1x128xi8> : i8
> >     %12 = "neura.add"(%9, %0) : (index, index) -> index
> >     %13 = "neura.cast"(%12) <{cast_type = "index_to_int"}> : (index) -> i64
> >     neura.br %13 : i64 to ^bb3
> > ^bb5:  // pred: ^bb3
> >     %14 = "neura.add"(%5, %0) : (index, index) -> index
> >     %15 = "neura.cast"(%14) <{cast_type = "index_to_int"}> : (index) -> i64
> >     neura.br %15 : i64 to ^bb1
> > ^bb6:  // pred: ^bb1
> >     "neura.return"() : () -> ()
> > }
> > ```
> 
> I believe we can `grant_predicate` further on the already `grant_always`, i.e., all the live-in of `bb2`, rather than the result of `bb2`.
> 
> * I think the rule is to (use `bb1` with `cond_br` as an example) `grant_predicate` all `bb1`'s live-out and all true-successor bb's (i.e., `bb2`) live-ins, WDYT? (and `grant_predicate(NOT)` on false-successor, i.e., `bb6`, though nothing to predicate for it as it has no argument in this IR example)
> * If one `const`/`grant_always` is used by multiple BBs being dominated by different `conditions`, we can `grant_predicate` them based on the specific `condition`
> * For now, we blindly give `grant_always` for all the const inside the entry block, we later can fuse the `grant_always` -> `grant_predicate`
> 
> How does this sound?

But there is still a problem, for `%2` in bb2, it

lib/NeuraDialect/Transforms/TransformCtrlToDataFlowPass.cpp
```diff
@@ -423,34 +425,61 @@ void createReserveAndPhiOps(func::FuncOp &func, ControlFlowInfo &ctrl_info,
...
-      llvm::errs()
-          << "[ctrl2data] Unsupported case: multiple conditions for a single block: "
-          << *target << "\n";
+      llvm::errs() << "[ctrl2data] Unsupported case: multiple conditions for a "
+                      "single block: "
+                   << *target << "\n";
       assert(false);
     }
 
     if (target->getArguments().empty()) {
```
tancheng, line 437: How do we cover the case where the live-ins contain both `args` and `directly used values from other blocks`?

ShangkunLi: For the kind of target blocks you mentioned, their incoming edges belong to Type 1, 2, 5, or 6. In these CFG dependency edge cases, we only need to use operations such as `phi`, `reserve`, `grant_predicate` to ensure the correctness of the incoming block argument (values passed through the edge), and the correctness of the other directly used live-in values can be guaranteed by SSA, so we don't need to consider them separately.

tancheng: Not sure we are on the same page, we can discuss during our weekly sync.

Basically, I am wondering that this `if (target->getArguments().empty())` is trying to handle the case where live-in is directly used (non-args) rather than passed in as args. So I am questioning whether we need to change this `if (target->getArguments().empty())` to `if (live_in_values != target->getArguments())`, my point is a `bb` could have live-ins of both args and non-args.

ShangkunLi: The code here is used to handle forward cond_br edges without values. And `if (target->getArguments().empty())` is used to double-check that the target block doesn't have block arguments. Because for this kind of target block, all the live-in values are defined in other blocks. We simply grant_predicate all the live-in values for such block.

As for blocks that have both block arguments and other live-in values. We directly use other live-in values. Because the predicate bit of the block argument can be passed to other operations in the block and ensure correctness. 

But to guarantee correctness, a better way is also to grant_predicate all the live-in values in the block (no matter if it is block arguments or other live-ins).

ShangkunLi: But I think for all the results in a block with both block arguments and other live-in values. The predicate bit of the arguments can be passed to other operations in this block, thus ensuring correctness. We do not have to grant_predicate other live-in values. For example,
```
^bb1(%param: !neura.data<f32, i1>): 
  %1 = "neura.fadd"(%param, %const1) : (!neura.data<f32, i1>, !neura.data<f32, i1>) -> !neura.data<f32, i1>
  %2 = "neura.fdiv"(%external1, %external2) : (!neura.data<f32, i1>, !neura.data<f32, i1>) -> !neura.data<f32, i1>
  %3 = "neura.fmul"(%1, %2) : (!neura.data<f32, i1>, !neura.data<f32, i1>) -> !neura.data<f32, i1>
```
Even though `%external1` and `%external2` are directly used live-in values. The correctness of this block can be guaranteed because the predicate bit is passed from `%param` -> `%1` -> `%3`.

ShangkunLi: Then I think we should grant_predicate all the other live_in values in the block.

tancheng: But my original intention is not `grant_predicate` all live-ins, I mean "grant specific predicate" to all live-ins in general, rather than targeting that specific neura operation/primitive.

> We cannot simply grant_predicate all the non-blockargument live-ins (e.g., %3, %2, %0). And there are also some cases that the incoming edges are all br, then it is impossible to grant_predicate live-ins in such cases because we don't even have a condition.

For such case, we need to find `%3, %2, %0`'s dominator bb:
 - if they come from a `br` w/o condition, we grant a `grant_always`
   - note that the `always` would be `and` with the %3's original predicate, rather than overwrite it
   - in this way, we don't need to specially handle entry-bb, which is normally embedded with `grant_once`
 - If they come from a `cond_br`, we grant a `grant_predicate` accordingly
   - again, the `grant_predicate` would be `and` with the %3's original predicate, rather than overwrite it

How does this sound? I don't know why this won't work.

tancheng: Ah right, I think this is legal from programmers perspective, but it is not that "legal" from Ctrl flow lowering in MLIR.

My original questions are about whether your `if block` can handle the common cases. Feel free to schedule meeting with me before Wed if you want.

test/neura/ctrl/branch_for.mlir
```diff
@@ -65,20 +65,20 @@ func.func @loop_test() -> f32 {
...
-// CTRL2DATA-NEXT:   %12 = neura.reserve : !neura.data<f32, i1>
-// CTRL2DATA-NEXT:   %13 = "neura.phi"(%9, %12) : (!neura.data<f32, i1>, !neura.data<f32, i1>) -> !neura.data<f32, i1>
-// CTRL2DATA-NEXT:   %14 = "neura.fadd"(%13, %7) : (!neura.data<f32, i1>, !neura.data<f32, i1>) -> !neura.data<f32, i1>
-// CTRL2DATA-NEXT:   %15 = "neura.add"(%11, %5) : (!neura.data<i64, i1>, !neura.data<i64, i1>) -> !neura.data<i64, i1>
+// CTRL2DATA-NEXT:   %10 = neura.reserve : !neura.data<{{f32|i64}}, i1>
+// CTRL2DATA-NEXT:   %11 = "neura.phi"(%10, %{{[0-9]+}}) : (!neura.data<{{f32|i64}}, i1>, !neura.data<{{f32|i64}}, i1>) -> !neura.data<{{f32|i64}}, i1>
+// CTRL2DATA-NEXT:   %12 = neura.reserve : !neura.data<{{f32|i64}}, i1>
+// CTRL2DATA-NEXT:   %13 = "neura.phi"(%12, %{{[0-9]+}}) : (!neura.data<{{f32|i64}}, i1>, !neura.data<{{f32|i64}}, i1>) -> !neura.data<{{f32|i64}}, i1>
+// CTRL2DATA-NEXT:   %14 = "neura.fadd"(%{{[0-9]+}}, %7) : (!neura.data<f32, i1>, !neura.data<f32, i1>) -> !neura.data<f32, i1>
+// CTRL2DATA-NEXT:   %15 = "neura.add"(%{{[0-9]+}}, %5) : (!neura.data<i64, i1>, !neura.data<i64, i1>) -> !neura.data<i64, i1>
```
tancheng, line 73: Why non-deterministic? Is it due to the `DenseSet` used inside the pass?

tancheng: And if you wanna use reg, plz label them with variable name to indicate the dependency.

test/neura/ctrl/branch_with_and_without_arg.mlir
```diff
@@ -49,24 +49,22 @@ func.func @test(%in: i64) -> f32 {
...
+// CTRL2DATA-NEXT:    %3 = "neura.constant"() <{predicate = true, value = 2.000000e+00 : f32}> : () -> !neura.data<f32, i1>
+// CTRL2DATA-NEXT:    %4 = "neura.grant_always"(%3) : (!neura.data<f32, i1>) -> !neura.data<f32, i1>
+// CTRL2DATA-NEXT:    %5 = "neura.grant_once"(%3) : (!neura.data<f32, i1>) -> !neura.data<f32, i1>
+// CTRL2DATA-NEXT:    %6 = "neura.constant"() <{predicate = true, value = 3.000000e+00 : f32}> : () -> !neura.data<f32, i1>
+// CTRL2DATA-NEXT:    %7 = "neura.grant_once"(%6) : (!neura.data<f32, i1>) -> !neura.data<f32, i1>
+// CTRL2DATA-NEXT:    %8 = "neura.icmp"(%arg0, %0) <{cmpType = "eq"}> : (i64, !neura.data<i64, i1>) -> !neura.data<i1, i1>
+// CTRL2DATA-NEXT:    %9 = "neura.grant_once"(%8) : (!neura.data<i1, i1>) -> !neura.data<i1, i1>
+// CTRL2DATA-DAG:    %{{[0-9]+}} = neura.grant_predicate %7, %9 : !neura.data<f32, i1>, !neura.data<i1, i1> -> !neura.data<f32, i1>
+// CTRL2DATA-DAG:    %{{[0-9]+}} = "neura.not"(%9) : (!neura.data<i1, i1>) -> !neura.data<i1, i1>
+// CTRL2DATA-DAG:    %{{[0-9]+}} = neura.grant_predicate %2, %{{[0-9]+}} : !neura.data<f32, i1>, !neura.data<i1, i1> -> !neura.data<f32, i1>
```
tancheng, line 64: I create this test specifically to test a `bb` has both arg and non-arg. Now, this is exactly the place you mentioned/assumed:
 - the non-arg live-in (e.g., `%2`) shouldn't be granted with any condition, as the live-outs (`%15`) would be granted with condition based on arg live-in (`%8`).

My above summary is correct?

test/neura/ctrl/branch_without_arg.mlir
```diff
@@ -51,25 +51,24 @@ func.func @test(%in: i64) -> f32 {
...
+// CTRL2DATA-NEXT:     %8 = "neura.grant_once"(%7) : (!neura.data<f32, i1>) -> !neura.data<f32, i1>
+// CTRL2DATA-NEXT:     %9 = "neura.icmp"(%arg0, %0) <{cmpType = "eq"}> : (i64, !neura.data<i64, i1>) -> !neura.data<i1, i1>
+// CTRL2DATA-NEXT:     %10 = "neura.grant_once"(%9) : (!neura.data<i1, i1>) -> !neura.data<i1, i1>
+// CTRL2DATA-DAG:      %11 = neura.grant_predicate %{{[0-9]+}}, %10 : !neura.data<f32, i1>, !neura.data<i1, i1> -> !neura.data<f32, i1>
+// CTRL2DATA-DAG:      %12 = neura.grant_predicate %{{[0-9]+}}, %10 : !neura.data<f32, i1>, !neura.data<i1, i1> -> !neura.data<f32, i1>
+// CTRL2DATA-NEXT:     %13 = "neura.not"(%10) : (!neura.data<i1, i1>) -> !neura.data<i1, i1>
+// CTRL2DATA-NEXT:     %14 = neura.grant_predicate %{{[0-9]+}}, %13 : !neura.data<f32, i1>, !neura.data<i1, i1> -> !neura.data<f32, i1>
+// CTRL2DATA-NEXT:     %15 = neura.grant_predicate %{{[0-9]+}}, %13 : !neura.data<f32, i1>, !neura.data<i1, i1> -> !neura.data<f32, i1>
+// CTRL2DATA-NEXT:     %16 = "neura.fadd"(%{{[0-9]+}}, %{{[0-9]+}}) : (!neura.data<f32, i1>, !neura.data<f32, i1>) -> !neura.data<f32, i1>
```
tancheng, line 70: Plz do not use `{{[0-9]+}}`, the best practice is to use reg symbol, but I am okay with %number. The reason is we need to capture the dependency in these tests.

tancheng, line 70: which two are `fadded`? This is the case for `br` w/o `cond`. This is an important case needs to be captured, we shouldn't hide it.

test/neura/ctrl/branch_for.mlir
ShangkunLi: > Why non-deterministic? Is it due to the `DenseSet` used inside the pass?

Yes. The `DenseSet` makes the value non-deterministic. I will use the reg symbol to denote the dependence.

tancheng: Can we use std::set to make it deterministic instead? Then you can change the reg back to %id.

test/neura/ctrl/branch_with_and_without_arg.mlir
ShangkunLi: Correct. I didn't grant_predicate `%2` because I assume the predicate bit can be passed through the arg `%8`.

tancheng: Can you then add such `assert(all live out are dominated by arg's predicate)` in the code? Or it already exists in your code?

Then I am wondering the `test/neura/ctrl/branch_without_arg.mlir`'s `fadd`.

ShangkunLi: This is the transformed ir:
```
module {
  func.func @test(%arg0: i64) -> f32 attributes {accelerator = "neura"} {
    %0 = "neura.constant"() <{predicate = true, value = 0 : i64}> : () -> !neura.data<i64, i1>
    %1 = "neura.constant"() <{predicate = true, value = 1.000000e+00 : f32}> : () -> !neura.data<f32, i1>
    %2 = "neura.grant_once"(%1) : (!neura.data<f32, i1>) -> !neura.data<f32, i1>
    %3 = "neura.constant"() <{predicate = true, value = 2.000000e+00 : f32}> : () -> !neura.data<f32, i1>
    %4 = "neura.grant_always"(%3) : (!neura.data<f32, i1>) -> !neura.data<f32, i1>
    %5 = "neura.grant_once"(%3) : (!neura.data<f32, i1>) -> !neura.data<f32, i1>
    %6 = "neura.constant"() <{predicate = true, value = 3.000000e+00 : f32}> : () -> !neura.data<f32, i1>
    %7 = "neura.grant_once"(%6) : (!neura.data<f32, i1>) -> !neura.data<f32, i1>
    %8 = "neura.icmp"(%arg0, %0) <{cmpType = "eq"}> : (i64, !neura.data<i64, i1>) -> !neura.data<i1, i1>
    %9 = "neura.grant_once"(%8) : (!neura.data<i1, i1>) -> !neura.data<i1, i1>
    %10 = neura.grant_predicate %7, %9 : !neura.data<f32, i1>, !neura.data<i1, i1> -> !neura.data<f32, i1>
    %11 = "neura.not"(%9) : (!neura.data<i1, i1>) -> !neura.data<i1, i1>
    %12 = neura.grant_predicate %5, %11 : !neura.data<f32, i1>, !neura.data<i1, i1> -> !neura.data<f32, i1>
    %13 = neura.grant_predicate %2, %11 : !neura.data<f32, i1>, !neura.data<i1, i1> -> !neura.data<f32, i1>
    %14 = "neura.fadd"(%13, %12) : (!neura.data<f32, i1>, !neura.data<f32, i1>) -> !neura.data<f32, i1>
    %15 = "neura.fmul"(%10, %4) : (!neura.data<f32, i1>, !neura.data<f32, i1>) -> !neura.data<f32, i1>
    %16 = "neura.phi"(%14, %15) : (!neura.data<f32, i1>, !neura.data<f32, i1>) -> !neura.data<f32, i1>
    "neura.return"(%16) : (!neura.data<f32, i1>) -> ()
  }
}
```

tancheng: I am talking about the tricky one: `test/neura/ctrl/branch_without_arg.mlir`'s `fadd`

test/neura/ctrl/branch_without_arg.mlir
ShangkunLi: Here it is
```
module {
  func.func @test(%arg0: i64) -> f32 attributes {accelerator = "neura"} {
    %0 = "neura.constant"() <{predicate = true, value = 0 : i64}> : () -> !neura.data<i64, i1>
    %1 = "neura.constant"() <{predicate = true, value = 1.000000e+00 : f32}> : () -> !neura.data<f32, i1>
    %2 = "neura.grant_always"(%1) : (!neura.data<f32, i1>) -> !neura.data<f32, i1>
    %3 = "neura.constant"() <{predicate = true, value = 2.000000e+00 : f32}> : () -> !neura.data<f32, i1>
    %4 = "neura.grant_always"(%3) : (!neura.data<f32, i1>) -> !neura.data<f32, i1>
    %5 = "neura.constant"() <{predicate = true, value = 3.000000e+00 : f32}> : () -> !neura.data<f32, i1>
    %6 = "neura.grant_once"(%5) : (!neura.data<f32, i1>) -> !neura.data<f32, i1>
    %7 = "neura.constant"() <{predicate = true, value = 4.000000e+00 : f32}> : () -> !neura.data<f32, i1>
    %8 = "neura.grant_once"(%7) : (!neura.data<f32, i1>) -> !neura.data<f32, i1>
    %9 = "neura.icmp"(%arg0, %0) <{cmpType = "eq"}> : (i64, !neura.data<i64, i1>) -> !neura.data<i1, i1>
    %10 = "neura.grant_once"(%9) : (!neura.data<i1, i1>) -> !neura.data<i1, i1>
    %11 = neura.grant_predicate %8, %10 : !neura.data<f32, i1>, !neura.data<i1, i1> -> !neura.data<f32, i1>
    %12 = neura.grant_predicate %6, %10 : !neura.data<f32, i1>, !neura.data<i1, i1> -> !neura.data<f32, i1>
    %13 = "neura.not"(%10) : (!neura.data<i1, i1>) -> !neura.data<i1, i1>
    %14 = neura.grant_predicate %4, %13 : !neura.data<f32, i1>, !neura.data<i1, i1> -> !neura.data<f32, i1>
    %15 = neura.grant_predicate %2, %13 : !neura.data<f32, i1>, !neura.data<i1, i1> -> !neura.data<f32, i1>
    %16 = "neura.fadd"(%15, %14) : (!neura.data<f32, i1>, !neura.data<f32, i1>) -> !neura.data<f32, i1>
    %17 = "neura.fmul"(%12, %11) : (!neura.data<f32, i1>, !neura.data<f32, i1>) -> !neura.data<f32, i1>
    %18 = "neura.phi"(%16, %17) : (!neura.data<f32, i1>, !neura.data<f32, i1>) -> !neura.data<f32, i1>
    "neura.return"(%18) : (!neura.data<f32, i1>) -> ()
  }
}
```

test/neura/ctrl/branch_with_and_without_arg.mlir
ShangkunLi: > Can you then add such `assert(all live out are dominated by arg's predicate)` in the code? Or it already exists in your code?
> 
> Then I am wondering the `test/neura/ctrl/branch_without_arg.mlir`'s `fadd`.

Add the assertion in the latest commit.

lib/NeuraDialect/Transforms/TransformCtrlToDataFlowPass.cpp
```diff
@@ -87,307 +102,549 @@ void GrantPredicateInEntryBlock(Block *entry_block, OpBuilder &builder) {
...
+    if (conditions.size() > 1) {
+      llvm::errs() << "[ctrl2data] Unsupported case: multiple conditions for a "
+                      "single block: "
+                   << *target << "\n";
+      assert(false);
+    }
+
+    if (target->getArguments().empty()) {
+      // Grants predicate for all the live-in values in the target block.
+      DenseSet<Value> live_in_values;
```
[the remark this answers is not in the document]
ShangkunLi: No. Changing it to llvm::denseVector, llvm::SmallVector, llvm::SetVector, and std::set didn't work. So finally, I chose to use the reg symbol in the test files.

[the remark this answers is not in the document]
ShangkunLi: added in latest commit.

test/neura/ctrl/branch_for.mlir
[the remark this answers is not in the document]
ShangkunLi: Sure, will enable in the next pr. I am coding on the ctrl-flow fusion now.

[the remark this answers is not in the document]
tancheng: Thanks for investigation.
 - We'd better avoid disabling tests, especially when there are more than one contributor, and some others are using the repo to ramp up
 - We prefer "tiny" PRs. One PR could only target single functionality or function, it might touch a lot of tests, which is fine, but the funtionality-wise, it should be "tiny"
   - If it is a large project, you could make a chain of branches, make PRs gradually

> --insert-data-mov can work and can generate correct intermediate ir

You mean the `intermediate ir` is exactly the same with or without your changes (this PR or your current implementation)?

> but the --map-to-accelerator doesn't work now

Hanging or crash or other issue?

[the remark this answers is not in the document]
tancheng: > Got it! Sorry for the problem caused by my unorthodox development process.

No worry :-)

> it will reach the maxII and exit without generating a legal mapping result.

It is probably due to unoptimized mapping strategy:
 - Only pick the lowest cost (highest award) tile to map and ignore all other candidate tiles: https://github.com/coredac/dataflow/blob/c58754e23fb1c4349aefd99054b107342e703167/lib/NeuraDialect/Mapping/mapping_util.cpp#L378-L390
 - No backtracking, i.e., if that placement or route failed, it just return false for that specific II, without back track to another tile and retry placement and route

Are you able to make a clean PR to just try to make the backtrackable mapping work to restore this test? No need to be super sophisticated though, https://github.com/coredac/dataflow/issues/59

PR #68, opened 2025-07-08
Fix run command syntax and check rules in nested_branch.mlir test

YanzhouTang:
## Description

The test was not properly validating output because the RUN command syntax was incorrect, causing it to falsely pass without actually running the transformation pipeline and FileCheck validation.

## Changes Made

1. Fix run command
2. Update CHECK rules in nested_branch.mlir test


tancheng (approved):
Thanks for the fix~!

PR #72, opened 2025-07-10
add architecture_spec.yaml file and yaml2json script

n0thingNoob:
Adding a architecture spec example and a simple yaml2json python script


tancheng:
I am thinking should we put these 3 files into a new `arch_specs` directory?

.gitignore
ShangkunLi, line 1: Kindly remind that please do not push your own `.gitignore` file into the repo. Thx~ ^_^

tancheng: Shouldn't this be fine? its rule can be applied to all users, as long as we know what we put inside this `ignore` file?

tancheng: What's the difference after adding this `.gitignore`? I thought it doesn't hurt? I normally use `git add new_file` before `git commit`. Why you need to keep your own local `.gitignore`?

ShangkunLi: Hmmm, I think we can maintain a better `.gitignore` file like https://github.com/llvm/llvm-project/blob/main/.gitignore#L57, so that we can simply use `git add .` before commit. And that's why I keep my own `.gitignore` before.

tancheng: got it, thanks for clarification. Feel free to remove this temporal `.gitignore` or fix it in another PR.

PR #85, opened 2025-07-19
[feature] Enable ctrl to data flow lowering for llvm.func

tancheng:
Enable ctrl to data flow lowering for llvm.func

### DIFF
--- a/lib/NeuraDialect/Transforms/TransformCtrlToDataFlowPass.cpp
+++ b/lib/NeuraDialect/Transforms/TransformCtrlToDataFlowPass.cpp
@@ -4,6 +4,7 @@
 #include "NeuraDialect/NeuraPasses.h"
 #include "NeuraDialect/NeuraTypes.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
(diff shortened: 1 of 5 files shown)


lib/NeuraDialect/Transforms/TransformCtrlToDataFlowPass.cpp
```diff
@@ -628,23 +630,42 @@ struct TransformCtrlToDataFlowPass
...
+      if (auto func = dyn_cast<func::FuncOp>(op)) {
+        auto accel_attr = func->getAttrOfType<StringAttr>("accelerator");
+        if (!accel_attr || accel_attr.getValue() != "neura") {
+          return;
+        }
+        region = &func.getBody();
+        domInfo = DominanceInfo(func);
+        GrantPredicateInEntryBlock(&region->front(), builder);
+        assertLiveOutValuesDominatedByBlockArgs(*region);
+      } else if (auto llvmFunc = dyn_cast<LLVM::LLVMFuncOp>(op)) {
```
ShangkunLi, line 652: Why not just transform such `llvm.func` operation into `func.func` operation in our `canonicalization` pass to avoid handling two different kinds of `func` operations?

ShangkunLi (approved):

PR #86, opened 2025-07-20
Create a neura canonicalize pass

ShangkunLi:
In this pr:

- promote live-in values in each bb to the block arguments
- promote the func arguments to `neura.constant`s so that we can predicate and map these func arguments


tancheng:
> Hi @tancheng, it seems we don't have a unified version of clang right now, which is why this test failed. Do we need to build the latest version of clang in the workflow? Or just sync the clang version to 14 as mentioned in #19.

Error msg:
```
/home/runner/work/dataflow/dataflow/test/c2llvm2mlir/Output/test.mlir.script: line 1: /home/runner/work/dataflow/dataflow/llvm-project/build/./bin/clang++: No such file or directory
```

Why these tests succeed on current master? There should be clang++ installed in github action, right?

ShangkunLi:
> > Hi @tancheng, it seems we don't have a unified version of clang right now, which is why this test failed. Do we need to build the latest version of clang in the workflow? Or just sync the clang version to 14 as mentioned in #19.
> 
> Error msg:
> 
> ```
> /home/runner/work/dataflow/dataflow/test/c2llvm2mlir/Output/test.mlir.script: line 1: /home/runner/work/dataflow/dataflow/llvm-project/build/./bin/clang++: No such file or directory
> ```
> 
> Why these tests succeed on current master? There should be clang++ installed in github action, right?

Because I changed the `test/lit.cfg.in` file to make tests use the clang built directly from source code, and I added the built scripts to build the clang in the latest commit.

In previous tests, `clang++` uses the clang installed in github action.

I prefer to use the clang built from the same source code as the mlir to keep consistency, WDYT? (Let's see how many hours it would take to build both clang & mlir and then make a decision)

test/neura/ctrl/branch_for.mlir
```diff
@@ -74,128 +77,197 @@ func.func @loop_test() -> f32 {
 // CHECK-NEXT:   "neura.return"(%10) : (!neura.data<f32, i1>) -> ()
 // CHECK-NEXT: }
 
-// CTRL2DATA:      func.func @loop_test() -> f32 attributes {accelerator = "neura"} {
+// CTRL2DATA:   func.func @loop_test() -> f32 attributes {accelerator = "neura"} {
```
tancheng, line 80: the additional `space` before `func.func` is to make it aligned with following `-NEXT: `.

test/neura/ctrl/branch_for.mlir
```diff
@@ -7,13 +7,15 @@
 // RUN: mlir-neura-opt %s \
 // RUN:   --assign-accelerator \
 // RUN:   --lower-llvm-to-neura \
+// RUN:   --neura-canonicalize \
 // RUN:   --leverage-predicated-value \
```
tancheng, line 11: Can we add another test before this, to only check the impact of `--neura-canonicalize`?

include/NeuraDialect/NeuraPasses.h
```diff
@@ -27,6 +27,7 @@ std::unique_ptr<mlir::Pass> createLeveragePredicatedValuePass();
 std::unique_ptr<mlir::Pass> createMapToAcceleratorPass();
 std::unique_ptr<mlir::Pass> createGenerateCodePass();
 std::unique_ptr<mlir::Pass> createFuseControlFlowPass();
+std::unique_ptr<mlir::Pass> createNeuraCanonicalizePass();
```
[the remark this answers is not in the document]
ShangkunLi: Good idea.

lib/Conversion/ArithToNeura/ArithToNeuraPass.cpp
```diff
@@ -39,7 +39,7 @@ struct ArithConstantToNeuraConstant
     Attribute value = op.getValue();
     // Optional predicate parameter can be null.
     rewriter.replaceOpWithNewOp<neura::ConstantOp>(op, result_type, value,
-                                                   nullptr);
+                                                   rewriter.getBoolAttr(true));
```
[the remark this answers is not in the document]
ShangkunLi: Maybe we don't need this for other `neura.operation`s, but we can at least add this attribute to `neura.constant` to indicate it is always valid.

BTW, I am preparing to simplify the logic of creating `grant_always` and `grant_once`. Since now we promote all the live-ins to the block args, we don't need such `grant_xxx` operations, or we can just embed the `grant_always` into the `neura.constant` ops.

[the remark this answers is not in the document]
tancheng: > we can at least add this attribute to `neura.constant` to indicate it is always valid.

That's my concern. We shouldn't need `valid = true` for constant. If it is not yet lowered to dataflow, it is purely ctrl flow, why do we need `valid = true`; if it is lowered to dataflow, the `data value` itself has `predicate`, we anyways don't need `valid = true` as an attribute.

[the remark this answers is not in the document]
ShangkunLi: Make sense. If we choose to use the `predicateDataType`, we can remove all the `predicate` attributes in the neura ops.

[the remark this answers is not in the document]
tancheng: We can handle this in another PR, plz file issue for this.

test/neura/ctrl/branch_for.mlir
ShangkunLi: Added in the latest commit.

lib/NeuraDialect/Transforms/CanonicalizeLiveInPass.cpp
```diff
@@ -0,0 +1,234 @@
...
+      if (auto br_op = dyn_cast<neura::Br>(pred_op)) {
+        if (br_op.getDest() == &block) {
+          // Creates a new operand list, including the original operands.
+          SmallVector<Value, 4> new_operands;
+
+          for (Value operand : br_op.getOperands()) {
+            new_operands.push_back(operand);
+          }
+
+          // Adds live-in values as new operands
```
tancheng, line 118: Period at the end.

lib/NeuraDialect/Transforms/CanonicalizeLiveInPass.cpp
```diff
@@ -0,0 +1,234 @@
...
+          for (Value live_in : live_ins) {
+            new_operands.push_back(live_in);
+          }
```
tancheng, line 121: What if the `live_in` is not declared inside this `pred BB`? (can we create a test for this case?)

lib/NeuraDialect/Transforms/CanonicalizeLiveInPass.cpp
```diff
@@ -0,0 +1,234 @@
...
+#define GEN_PASS_DEF_NEURACANONICALIZE
+#include "NeuraDialect/NeuraPasses.h.inc"
+
+namespace {
+LogicalResult promoteFunctionArgsToConstants(Region &region) {
+  if (region.empty()) {
+    return success();
+  }
+
+  Block &entryBlock = region.front();
```
tancheng, line 26: `entryBlock` -> `entry_block`

lib/NeuraDialect/Transforms/CanonicalizeLiveInPass.cpp
```diff
@@ -0,0 +1,234 @@
...
+  OpBuilder builder(&entryBlock, entryBlock.begin());
+
+  // Collects all function arguments.
+  SmallVector<BlockArgument, 4> args(entryBlock.getArguments().begin(),
+                                     entryBlock.getArguments().end());
+
+  // Creates a constant operation for each function argument.
+  for (auto [idx, arg] : llvm::enumerate(args)) {
+    // For constant operation, the default predicate is true.
+    auto constOp = builder.create<neura::ConstantOp>(
```
tancheng, line 36: `constOp` -> `const_op`

tancheng:
> > > Hi @tancheng, it seems we don't have a unified version of clang right now, which is why this test failed. Do we need to build the latest version of clang in the workflow? Or just sync the clang version to 14 as mentioned in #19.
> > 
> > 
> > Error msg:
> > ```
> > /home/runner/work/dataflow/dataflow/test/c2llvm2mlir/Output/test.mlir.script: line 1: /home/runner/work/dataflow/dataflow/llvm-project/build/./bin/clang++: No such file or directory
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Why these tests succeed on current master? There should be clang++ installed in github action, right?
> 
> Because I changed the `test/lit.cfg.in` file to make tests use the clang built directly from source code, and I added the built scripts to build the clang in the latest commit.
> 
> In previous tests, `clang++` uses the clang installed in github action.
> 
> I prefer to use the clang built from the same source code as the mlir to keep consistency, WDYT? (Let's see how many hours it would take to build both clang & mlir and then make a decision)

3h looks fine to me.. Then what should I do on my local docker? Do I need to install a specific llvm/clang version to build mlir? Or do i need that specific llvm/clang to run our test?

ShangkunLi:
> > > > Hi @tancheng, it seems we don't have a unified version of clang right now, which is why this test failed. Do we need to build the latest version of clang in the workflow? Or just sync the clang version to 14 as mentioned in #19.
> > > 
> > > 
> > > Error msg:
> > > ```
> > > /home/runner/work/dataflow/dataflow/test/c2llvm2mlir/Output/test.mlir.script: line 1: /home/runner/work/dataflow/dataflow/llvm-project/build/./bin/clang++: No such file or directory
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Why these tests succeed on current master? There should be clang++ installed in github action, right?
> > 
> > 
> > Because I changed the `test/lit.cfg.in` file to make tests use the clang built directly from source code, and I added the built scripts to build the clang in the latest commit.
> > In previous tests, `clang++` uses the clang installed in github action.
> > I prefer to use the clang built from the same source code as the mlir to keep consistency, WDYT? (Let's see how many hours it would take to build both clang & mlir and then make a decision)
> 
> 3h looks fine to me.. Then what should I do on my local docker? Do I need to install a specific llvm/clang version to build mlir? Or do i need that specific llvm/clang to run our test?

Use this cmake cripts to build the clang. And then use the `clang++` in `build/bin/clang++` instead.

```
git clone https://github.com/llvm/llvm-project.git
cd llvm-project
git checkout 6146a88
mkdir build && cd build
cmake -G Ninja ../llvm \
-DLLVM_ENABLE_PROJECTS="mlir;clang" \
-DLLVM_BUILD_EXAMPLES=OFF \
-DLLVM_TARGETS_TO_BUILD="Native" \
-DCMAKE_BUILD_TYPE=Release \
-DLLVM_ENABLE_ASSERTIONS=ON \
-DCMAKE_C_COMPILER=clang \
-DCMAKE_CXX_COMPILER=clang++ \
-DCMAKE_CXX_STANDARD=17 \
-DCMAKE_CXX_FLAGS="-std=c++17 -frtti" \
-DLLVM_ENABLE_LLD=ON \
-DMLIR_INSTALL_AGGREGATE_OBJECTS=ON \
-DLLVM_ENABLE_RTTI=ON \
-DCMAKE_C_COMPILER_LAUNCHER=ccache \
-DCMAKE_CXX_COMPILER_LAUNCHER=ccache

cmake --build . --target check-mlir
cmake --build . --target check-clang
```

tancheng:
> > > > > Hi @tancheng, it seems we don't have a unified version of clang right now, which is why this test failed. Do we need to build the latest version of clang in the workflow? Or just sync the clang version to 14 as mentioned in #19.
> > > > 
> > > > 
> > > > Error msg:
> > > > ```
> > > > /home/runner/work/dataflow/dataflow/test/c2llvm2mlir/Output/test.mlir.script: line 1: /home/runner/work/dataflow/dataflow/llvm-project/build/./bin/clang++: No such file or directory
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > Why these tests succeed on current master? There should be clang++ installed in github action, right?
> > > 
> > > 
> > > Because I changed the `test/lit.cfg.in` file to make tests use the clang built directly from source code, and I added the built scripts to build the clang in the latest commit.
> > > In previous tests, `clang++` uses the clang installed in github action.
> > > I prefer to use the clang built from the same source code as the mlir to keep consistency, WDYT? (Let's see how many hours it would take to build both clang & mlir and then make a decision)
> > 
> > 
> > 3h looks fine to me.. Then what should I do on my local docker? Do I need to install a specific llvm/clang version to build mlir? Or do i need that specific llvm/clang to run our test?
> 
> Use this cmake cripts to build the clang. And then use the `clang++` in `build/bin/clang++` instead.
> 
> ```
> git clone https://github.com/llvm/llvm-project.git
> cd llvm-project
> git checkout 6146a88
> mkdir build && cd build
> cmake -G Ninja ../llvm \
> -DLLVM_ENABLE_PROJECTS="mlir;clang" \
> -DLLVM_BUILD_EXAMPLES=OFF \
> -DLLVM_TARGETS_TO_BUILD="Native" \
> -DCMAKE_BUILD_TYPE=Release \
> -DLLVM_ENABLE_ASSERTIONS=ON \
> -DCMAKE_C_COMPILER=clang \
> -DCMAKE_CXX_COMPILER=clang++ \
> -DCMAKE_CXX_STANDARD=17 \
> -DCMAKE_CXX_FLAGS="-std=c++17 -frtti" \
> -DLLVM_ENABLE_LLD=ON \
> -DMLIR_INSTALL_AGGREGATE_OBJECTS=ON \
> -DLLVM_ENABLE_RTTI=ON \
> -DCMAKE_C_COMPILER_LAUNCHER=ccache \
> -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
> 
> cmake --build . --target check-mlir
> cmake --build . --target check-clang
> ```

Plz update README.md then.

lib/NeuraDialect/Transforms/CanonicalizeLiveInPass.cpp
tancheng: This seems to imply that we need to iterate the BB in topological ordering; otherwise, the `pred BB` could have no declared live-in. Can we have a test like this?

ShangkunLi: This test case has been added in `test/neura/ctrl/nested_branch.mlir`. The bb-sort function is added in the latest commit.

lib/NeuraDialect/Transforms/CanonicalizeLiveInPass.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+  std::function<void(Block *)> dfs = [&](Block *block) {
+    visited.insert(block);
+
+    Operation *terminator = block->getTerminator();
+    if (terminator) {
+      for (unsigned i = 0; i < terminator->getNumSuccessors(); ++i) {
+        Block *succ = terminator->getSuccessor(i);
+        if (!visited.count(succ)) {
+          dfs(succ);
```
tancheng, line 39: Why DFS w/ reverse at the end, instead of BFS w/o reverse? Either might be fine, just curious about your motivation.

And do you think it is necessary to have a test for this?

ShangkunLi: Hmmmm, I don't think we need such a test. If we do not use the canonicalize pass, we don't need to consider such a case. Conversely, if we use the canonicalize pass, the sort function will ensure the correctness directly.

And the above-mentioned case may not fail because if we directly use `for(Block& block: region.getBlocks())` to iterate over each block. The order is not visible to us unless we run the pass, hence, it is a little bit tricky to create such a case.

tancheng: > if we use the canonicalize pass, the sort function will ensure the correctness directly.

This is what we want to test. Theoretically, all the new functionality needs a corresponding unit test. I don't get `the above-mentioned case may not fail because if we directly use for(Block& block: region.getBlocks()) to iterate over each block.` I thought this would by default iterates `entry bb` -> `bb1` -> `bb2` -> `bb3`, which would fail w/o your new topological ordering, no?

ShangkunLi: The order indeed is `entry bb` -> `bb1` -> `bb2` -> `bb3`. But it will not fail, because SSA guarantees that each live-in is defined in the previous bbs when we try to handle a bb. 

For example,
```
  llvm.func @control_flow_example(%arg0: i32) -> i32 {
    %0 = llvm.mlir.constant(1 : i32) : i32
    %1 = llvm.mlir.constant(2 : i32) : i32
    %2 = llvm.mlir.constant(10 : i32) : i32
    llvm.br ^bb2

  ^bb1(%x: i32):
    llvm.return %x : i32

  ^bb2:
    %3 = llvm.add %arg0, %0 : i32
    %4 = llvm.icmp "slt" %3, %2 : i32
    llvm.cond_br %4, ^bb1(%1 : i32), ^bb3

  ^bb3:
    %5 = llvm.add %3, %1 : i32
    llvm.br ^bb1(%5 : i32)
  }
```

The canonicalized ir is (transformed without sorting)
```
  llvm.func @control_flow_example(%arg0: i32) -> i32 attributes {accelerator = "neura"} {
    %0 = "neura.constant"() <{predicate = true, value = "%arg0"}> : () -> !neura.data<i32, i1>
    %1 = "neura.constant"() <{predicate = true, value = 1 : i32}> : () -> !neura.data<i32, i1>
    %2 = "neura.constant"() <{predicate = true, value = 2 : i32}> : () -> !neura.data<i32, i1>
    %3 = "neura.constant"() <{predicate = true, value = 10 : i32}> : () -> !neura.data<i32, i1>
    neura.br %0, %1, %3, %2 : !neura.data<i32, i1>, !neura.data<i32, i1>, !neura.data<i32, i1>, !neura.data<i32, i1> to ^bb2
  ^bb1(%4: !neura.data<i32, i1>):  // 2 preds: ^bb2, ^bb3
    "neura.return"(%4) : (!neura.data<i32, i1>) -> ()
  ^bb2(%5: !neura.data<i32, i1>, %6: !neura.data<i32, i1>, %7: !neura.data<i32, i1>, %8: !neura.data<i32, i1>):  // pred: ^bb0
    %9 = "neura.add"(%5, %6) : (!neura.data<i32, i1>, !neura.data<i32, i1>) -> !neura.data<i32, i1>
    %10 = "neura.icmp"(%9, %7) <{cmpType = "slt"}> : (!neura.data<i32, i1>, !neura.data<i32, i1>) -> !neura.data<i1, i1>
    neura.cond_br %10 : !neura.data<i1, i1> then %8 : !neura.data<i32, i1> to ^bb1 else %9, %2 : !neura.data<i32, i1>, !neura.data<i32, i1> to ^bb3
  ^bb3(%11: !neura.data<i32, i1>, %12: !neura.data<i32, i1>):  // pred: ^bb2
    %13 = "neura.add"(%11, %12) : (!neura.data<i32, i1>, !neura.data<i32, i1>) -> !neura.data<i32, i1>
    neura.br %13 : !neura.data<i32, i1> to ^bb1
  }
```

ShangkunLi: For the source ir
```
  llvm.func @control_flow_example(%arg0: i32) -> i32 {
    %0 = llvm.mlir.constant(1 : i32) : i32
    %1 = llvm.mlir.constant(2 : i32) : i32
    %2 = llvm.mlir.constant(10 : i32) : i32
    %6 = llvm.mlir.constant(1 : i1) : i1
    llvm.br ^bb2

  ^bb1(%x: i32):
    llvm.return %x : i32

  ^bb2:
    %3 = llvm.add %arg0, %0 : i32
    %4 = llvm.icmp "slt" %3, %2 : i32
    llvm.cond_br %6, ^bb1(%1 : i32), ^bb3

  ^bb3:
    %5 = llvm.add %3, %1 : i32
    llvm.br ^bb1(%5 : i32)
  }
```

We can still transform it into the ir below without sorting because of the SSA.
```
  llvm.func @control_flow_example(%arg0: i32) -> i32 attributes {accelerator = "neura"} {
    %0 = "neura.constant"() <{predicate = true, value = "%arg0"}> : () -> !neura.data<i32, i1>
    %1 = "neura.constant"() <{predicate = true, value = 1 : i32}> : () -> !neura.data<i32, i1>
    %2 = "neura.constant"() <{predicate = true, value = 2 : i32}> : () -> !neura.data<i32, i1>
    %3 = "neura.constant"() <{predicate = true, value = true}> : () -> !neura.data<i1, i1>
    neura.br %0, %1, %3, %2 : !neura.data<i32, i1>, !neura.data<i32, i1>, !neura.data<i1, i1>, !neura.data<i32, i1> to ^bb2
  ^bb1(%4: !neura.data<i32, i1>):  // 2 preds: ^bb2, ^bb3
    "neura.return"(%4) : (!neura.data<i32, i1>) -> ()
  ^bb2(%5: !neura.data<i32, i1>, %6: !neura.data<i32, i1>, %7: !neura.data<i1, i1>, %8: !neura.data<i32, i1>):  // pred: ^bb0
    %9 = "neura.add"(%5, %6) : (!neura.data<i32, i1>, !neura.data<i32, i1>) -> !neura.data<i32, i1>
    neura.cond_br %7 : !neura.data<i1, i1> then %8 : !neura.data<i32, i1> to ^bb1 else %9, %2 : !neura.data<i32, i1>, !neura.data<i32, i1> to ^bb3
  ^bb3(%10: !neura.data<i32, i1>, %11: !neura.data<i32, i1>):  // pred: ^bb2
    %12 = "neura.add"(%10, %11) : (!neura.data<i32, i1>, !neura.data<i32, i1>) -> !neura.data<i32, i1>
    neura.br %12 : !neura.data<i32, i1> to ^bb1
  }
```

ShangkunLi: Agree!

PR #88, opened 2025-07-20
Code Generate Pass

n0thingNoob:
## Current work
- Fixed source direction calculation for operations with data_mov operands
- Added op_to_source_tile mapping to track original source locations
- Enhanced ASM generation with correct Entry conditions and operand directions
- Added test results in the Generated_Code folder

## Future Enhancements Needed
 - Add comprehensive error handling and validation
 - Refactor duplicate code in JSON/ASM generation
 - Add configuration options for output customization
 - Implement unit tests and integration tests
 - Support more complex data flow patterns
 - Add debugging and diagnostic capabilities

## Testing
 - Tested with test/test.mlir, the results of asm and json look right, but might be missing some details.

## Files Changed
 - lib/NeuraDialect/Transforms/GenerateCodePass.cpp
 - Generated_Code/ (new folder with test results)


lib/NeuraDialect/Transforms/GenerateCodePass.cpp
```diff
@@ -47,84 +72,691 @@ struct GenerateCodePass
...
+                }
+              }
+            }
+          }
+        }
+        
+        // Write Entry line
+        asm_out << "    Entry ";
+        bool first = true;
+        for (const auto& dir : input_directions) {
```
tancheng, line 611: ditto

lib/NeuraDialect/Transforms/GenerateCodePass.cpp
```diff
@@ -47,84 +72,691 @@ struct GenerateCodePass
...
+        if (time_ops.size() > 1) {
+          asm_out << "Loop {\n";
+        } else {
+          asm_out << "Once {\n";
+        }
+        
+        // Generate operations for each time step
+        for (auto& time_pair : time_ops) {
+          int time_step = time_pair.first;
+          auto& ops = time_pair.second;
```
tancheng, line 631: ditto

lib/NeuraDialect/Transforms/GenerateCodePass.cpp
```diff
@@ -47,84 +72,691 @@ struct GenerateCodePass
...
+        
+        // Determine if it's Loop or Once based on time steps
+        if (time_ops.size() > 1) {
+          asm_out << "Loop {\n";
+        } else {
+          asm_out << "Once {\n";
+        }
+        
+        // Generate operations for each time step
+        for (auto& time_pair : time_ops) {
```
tancheng, line 629: ditto

Generated_Code/generated-instructions.asm
tancheng, line 1: Shouldn't the `.md`, `.asm`, and `.json`, be located within the `test` folder?

lib/NeuraDialect/Transforms/GenerateCodePass.cpp
```diff
@@ -47,84 +72,691 @@ struct GenerateCodePass
...
+      std::map<Operation*, std::pair<int, int>> op_to_source_tile;
+      
+      // map of link ID to source and destination tiles
+      std::map<int, std::pair<std::pair<int, int>, std::pair<int, int>>> link_to_tiles;
+      
+      // map of PE coordinates to ID
+      std::map<std::pair<int, int>, int> pe_coord_to_id;
+      int pe_id_counter = 0;
 
+      // first pass: collect all operation tile locations and assign PE ID
```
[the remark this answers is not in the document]
n0thingNoob: Sure

Generated_Code/test.mlir
```diff
@@ -0,0 +1,12 @@
+// RUN: mlir-neura-opt %s | FileCheck %s
```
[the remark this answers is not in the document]
n0thingNoob: Yes, we need that option to generate the instruction.asm. We can generate json and asm at the same time.

Generated_Code/test.mlir
```diff
@@ -0,0 +1,12 @@
...
+func.func @test() -> f32 {
+  %a = arith.constant 1.0 : f32
+  %b = arith.constant 2.0 : f32
+  %res = "neura.fadd" (%a, %b) : (f32, f32) -> f32
+
+// Checks the expected lowered operation.
+// CHECK: neura.fadd
+
+  return %res : f32
+}
```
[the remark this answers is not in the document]
n0thingNoob: I am not sure how to check the generated code here.

ShangkunLi:
> @ShangkunLi any idea why the build time out? no changes related to mapper in this PR.

@n0thingNoob @tancheng I built and ran the code in this branch locally, the segment faults seem to come from the `--generate-code` pass.

n0thingNoob:
> > > @ShangkunLi any idea why the build time out? no changes related to mapper in this PR.
> > 
> > 
> > @n0thingNoob @tancheng I built and ran the code in this branch locally, the segment faults seem to come from the `--generate-code` pass.
> 
> ok, let me check what happened.

@ShangkunLi How can I run the same test locally? I want to see what happened to my code

tancheng:
@ShangkunLi plz help review this PR, especially `CodeGeneratePass`.

lib/NeuraDialect/Transforms/GenerateCodePass.cpp
```diff
@@ -14,118 +24,1364 @@ using namespace mlir::neura;
...
+      func_obj["ResMII"] = resMII_attr.getInt();
 
-      func.walk([&](Operation *op) {
-        if (isa<func::ReturnOp>(op))
-          return;
+    // Set maps instructions organized by tile.
+    std::map<std::pair<int, int>, YAMLNode> tile_instructions;
+    
+    // Set maps operation to its mapping location, for calculating data movement direction.
+    std::map<Operation*, std::pair<int, int>> op_to_tile;
```
ShangkunLi, line 596: I don't get where you use this `op_to_tile` variable. It seems like you're just assigning values to `op_to_tiles`, you're not really using it.

generated-instructions.yaml
```diff
@@ -0,0 +1,629 @@
...
+              - neura.data_mov
+            result_types: 
+              - !neura.data<f32, i1>
+            src_direction: Local
+            src_tile: (3,1)
+            time_step: 3
+        x: 3
+        y: 1
+      Tile(7): 
+        id: 7
```
ShangkunLi, line 365: It seems that you are wrongly calculating the tile id or tile coordinate. The coordinate of tile#7 should be (3,1) according to the mapping result.

test/codegenerate/generated-instructions.json
ShangkunLi, line 1: Maybe we should remove this file from being committed, because the `--generate-code` can generate it automatically, right?

generated-instructions.yaml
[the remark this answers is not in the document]
n0thingNoob: Just for testing, I will change it later.

test/mapped.mlir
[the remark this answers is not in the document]
n0thingNoob: This is a mistakenly updated file. Not the test file.

lib/NeuraDialect/Transforms/GenerateCodePass.cpp
```diff
@@ -62,9 +64,100 @@ struct GenerateCodePass
 
   StringRef getArgument() const override { return "generate-code"; }
   StringRef getDescription() const override {
-    return "Generates JSON code from mapped Neura IR.";
+    return "Generates YAML code from mapped Neura IR.";
   }
 
+  // Simple YAML node class
+  struct YAMLNode {
```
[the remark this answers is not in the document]
n0thingNoob: This class is specifically for building and representing YAML format data structures. It represents a node in a YAML document.

Actually, I cannot set the yamlcpp in my environment. And I have to use the sstream to generate the yaml.

lib/NeuraDialect/Transforms/CMakeLists.txt
```diff
@@ -25,4 +25,5 @@ add_mlir_library(
       MLIRTransforms
       MLIRNeura
       ${dialect_libs}
+      LLVMSupport
```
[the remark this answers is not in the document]
n0thingNoob: For String handling, File I/O and debug output, For example:
llvm::StringRef fullOpName = op->getName().getStringRef();
llvm::raw_fd_ostream asm_out("generated-instructions.asm", asm_ec);
llvm::errs() << "DEBUG: asm_op_to_final_tile size: " << asm_op_to_final_tile.size() << "\n";

But I can replace it with the c++ std lib.

test/codegenerate/generated-instructions.json
n0thingNoob: ok

generated-instructions.yaml
n0thingNoob: I have fixed it.

n0thingNoob:
I’ve rebuilt the code; it now uses the following methods to determine routing directions:

Build topology from the Architecture. For each link, use getSrcTile() / getDstTile() to record its endpoints and each tile’s (x, y).

Derive link direction. Compute the cardinal direction (EAST/WEST/NORTH/SOUTH) from the endpoints’ Δx/Δy, and provide invertDir for the opposite.

Read link chains from IR. For each producer→consumer edge, obtain the link chain from the consumer-side forwarder’s mapping_locs (resource="link"), sorted by time_step.

Endpoint wiring.

Producer dst = direction of the first link in the chain.

Consumer src = invert(direction of the last link).

Materialize multi-hop routing. If the chain has ≥ 2 hops, insert per-hop DATA_MOV on srcTileOfLink(cur) at the hop’s timestep, with
src = invert(direction(prev)) and dst = direction(cur) (deduplicated by (tile, ts, link)).

Same-tile edges. Use a local register $N (prefer mapped IDs; otherwise allocate per-tile in a stable manner).

ctrl_mov (loop-carried). Map reserve → phi, then apply the same rules using the ctrl_mov link chain; prefer its mapped register.

Fallback. If no link steps are present, infer directions from the Manhattan delta between producer and consumer tiles.

Output. Emit YAML/ASM unchanged; only multi-hop router hops are materialized as DATA_MOV.

lib/NeuraDialect/Transforms/GenerateCodePass.cpp
```diff
@@ -1,138 +1,829 @@
...
+      Operation *tp = chaseToTiledProducer(producer, opPlace);
+      auto pp = opPlace.lookup(tp);
+      rn = allocTileReg({pp.col, pp.row});
+    }
+    localRegOf[root] = rn;
+    return rn;
+  }
+
+  // map of tile(x,y) -> time_step -> vector<Instruction>
+  std::map<std::pair<int,int>, std::map<int, std::vector<Instruction>>> pe_time_insts;
```
tancheng, line 355: What do you mean by "pe"?

test/code_gen/test_code_generate.mlir
```diff
@@ -0,0 +1,35 @@
...
+// CHECK-DAG: "name": "neura.fadd"
+// CHECK-DAG: "neura.return"
+// CHECK-DAG: "CompiledII": 1
+// CHECK-DAG: "ResMII": 1
+// CHECK-DAG: "dst_direction": "East"
+// CHECK-DAG: "src_direction": "Local"
+// CHECK-DAG: "tile_instructions"
+// CHECK-DAG: "func_name": "test"
+
+// ASM: PE(0,2):
```
tancheng, line 23: Plz provide the entire ASM for this test, and CHECK ASM for all lines, so we get better idea about what we are generating.

tancheng:
Thanks @n0thingNoob for the update,
```
4. Endpoint wiring
Producer dst = direction of the first link in the chain.

Consumer src = invertDir(direction of the last link).

5. Multi-hop materialization (only if chain length ≥ 2)
For consecutive links (prev, cur) at time ts, place a router hop on srcTileOfLink(cur):

src = invertDir(dirFromLink(prev))

dst = dirFromLink(cur)

De-duplicate by (tile, ts, link).

6. Same-tile edges
Use a local register $N (prefer mapped IDs; otherwise allocate per-tile in a stable round-robin).

Attach $N as producer dst and consumer src.
```
Can you plz elaborate on above 4/5/6?
 - Why we need `invertDir()`?
 - What does "De-duplicate" do?
 - I thought we provide all needed registers, why you need to assign again by youself? CCing @ShangkunLi here.

lib/NeuraDialect/Transforms/GenerateCodePass.cpp
```diff
@@ -1,138 +1,829 @@
...
+
+struct Entry {
+  std::string entry_id;
+  std::string type;
+  std::vector<Instruction> instructions;
+  Entry(const std::string &id, const std::string &t = "loop")
+      : entry_id(id), type(t) {}
+};
+
+struct Core {
```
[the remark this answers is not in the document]
n0thingNoob: The format that @Jackcuii provided is using core, but we can discuss this later.

n0thingNoob:
Updates @tancheng 
- Naming cleanup: Standardized variable and function names(still need review).
- Local-tile register assignment: Removed random/round-robin allocation. Same-tile, cross-step edges (incl. ctrl_mov) now must use the register ID from mapping_locs; we error + assert if missing. Consumers are rewired to the exact $id.
- Tests updated: Refreshed MLIR tests (Used the branch_for) and YAML/ASM to reflect fixed register IDs and full dumps.

lib/NeuraDialect/Transforms/GenerateCodePass.cpp
```diff
@@ -1,138 +1,829 @@
...
+      if (!d) continue;
+      auto res = dyn_cast_or_null<StringAttr>(d.get("resource"));
+      if (!res) continue;
+      if (res.getValue() == "register" || res.getValue() == "reg") {
+        if (auto id = dyn_cast_or_null<IntegerAttr>(d.get("id")))
+          return id.getInt();
+      }
+    }
+  }
+  return std::nullopt;
```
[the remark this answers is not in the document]
n0thingNoob: Already solved in the new updates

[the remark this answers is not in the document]
tancheng: Can't we add `assert()` here?

lib/NeuraDialect/Transforms/GenerateCodePass.cpp
```diff
@@ -1,138 +1,803 @@
...
+    if (auto integer_attr = dyn_cast<IntegerAttr>(value_attr))
+      return "#" + std::to_string(integer_attr.getInt());
+    if (auto float_attr = dyn_cast<FloatAttr>(value_attr))
+      return "#" + std::to_string(float_attr.getValueAsDouble());
+  }
+  return "#0";
+}
+
+// Chase up the operation chain to find the nearest ancestor that has been assigned to a tile.
+static Operation *chaseToTiledProducer(Operation *op,
```
tancheng, line 152: Change to `getFirstAncestorMappedOnTile()`

lib/NeuraDialect/Transforms/GenerateCodePass.cpp
```diff
@@ -1,138 +1,803 @@
...
+        if (!register_id || !time_step) continue;
+        steps.push_back({(int)register_id.getInt(), (int)time_step.getInt()});
+      }
+    }
+  }
+  llvm::sort(steps, [](const RegStep &a, const RegStep &b){ return a.ts < b.ts; });
+  return steps;
+}
+
+// Only check for register id on the last forwarder(op) at consumer side or on op itself.
```
tancheng, line 254: `// Only checks`
`forwarder(op)` -> `forwarder (op)`

lib/NeuraDialect/Transforms/GenerateCodePass.cpp
```diff
@@ -1,138 +1,803 @@
...
+static std::optional<int> regOnBackwardPath(Value vAtConsumer) {
```
tancheng, line 255: `vAtConsumer` -> `value_of_consumer`

lib/NeuraDialect/Transforms/GenerateCodePass.cpp
```diff
@@ -295,110 +296,566 @@ struct GenerateCodePass
...
+        }
+      }
+    }
+    if (!rewired_to_register) setConsumerSourceExact(consumer, value_at_consumer, consumer_direction);
+  }
+
+  void wireEdges(func::FuncOp function, const Topology &topology) {
+    DenseMap<Operation*, ProducerSummary> producer_summary;
+    
+    function.walk([&](Operation *producer){
```
tancheng, line 463: Can you please comment in the caller that what you are doing using these helper functions. Why we need 6 times of `function.walk()` to construct what (can't we just walk once and get what we need?).

n0thingNoob: Yes, I have already made the change.

tancheng (commented):
Great job Yijia~!

Plz fix the comments style and variable naming convention. Also, plz let me click on the "solve" once you fix all of them. Thanks.

n0thingNoob:
> Great job Yijia~!
> 
> Plz fix the comments style and variable naming convention. Also, plz let me click on the "solve" once you fix all of them. Thanks.

Will do it ASAP.

test/code_gen/test_code_generate.mlir
```diff
@@ -0,0 +1,596 @@
...
+// YAML-NEXT:     - column: 0
+// YAML-NEXT:       row: 0
+// YAML-NEXT:       core_id: "0"
+// YAML-NEXT:       entries:
+// YAML-NEXT:         - entry_id: "entry0"
+// YAML-NEXT:           instructions:
+// YAML-NEXT:             - opcode: "CONSTANT"
+// YAML-NEXT:               timestep: 0
+// YAML-NEXT:               src_operands:
+// YAML-NEXT:                 - operand: "#0"
```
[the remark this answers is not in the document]
n0thingNoob: I didn't separate the constant value and register value inside the operand. The value starts with # means immediate number value, and the value starts with $ is the reg value.

test/code_gen/test_code_generate.mlir
```diff
@@ -0,0 +1,596 @@
...
+// YAML-NEXT:                 - operand: "NORTH"
+// YAML-NEXT:                   color: "RED"
+// YAML-NEXT:                 - operand: "$22"
+// YAML-NEXT:                   color: "RED"
+// YAML-NEXT:                 - operand: "$21"
+// YAML-NEXT:                   color: "RED"
+// YAML-NEXT:                 - operand: "SOUTH"
+// YAML-NEXT:                   color: "RED"
+// YAML-NEXT:                 - operand: "$20"
+// YAML-NEXT:                   color: "RED"
```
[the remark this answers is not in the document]
n0thingNoob: Yes, in the mapping IR, the ICMP produces a value ( %38).
%38 = "neura.icmp"(%36, %37) <{cmpType = "slt"}> {mapping_locs = [{id = 5 : i32, resource = "tile", time_step = 4 : i32, x = 1 : i32, y = 1 : i32}]} : (!neura.data<i64, i1>, !neura.data<i64, i1>) -> !neura.data<i1, i1>
I remember this %38 has been used by 6 data_mov and that's why there are 6 output operands.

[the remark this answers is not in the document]
tancheng: Thanks for clarification. Dumping the mapping IR would help us to understand/double-check the generated instructions. Plz include it~ Thanks.

test/code_gen/test_code_generate.mlir
```diff
@@ -0,0 +1,596 @@
...
+// YAML-NEXT:               dst_operands:
+// YAML-NEXT:                 - operand: "NORTH"
+// YAML-NEXT:                   color: "RED"
+// YAML-NEXT:                 - operand: "$4"
+// YAML-NEXT:                   color: "RED"
+// YAML-NEXT:         - entry_id: "entry3"
+// YAML-NEXT:           instructions:
+// YAML-NEXT:             - opcode: "DATA_MOV"
+// YAML-NEXT:               timestep: 3
```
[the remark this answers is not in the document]
tancheng: Thanks for the clarification. Similarly, this constraint/limitation needs to be explained in your README.md (not sure why you remove README.md in the latest commit though).

n0thingNoob:
I am not sure why there is a conflict when pushing the first commit. I merged the latest main into architecture_spec to resolve it and removed the obsolete README that was only used by the old test. @tancheng

tancheng:
> I am not sure why there is a conflict when pushing the first commit. I merged the latest main into architecture_spec to resolve it and removed the obsolete README that was only used by the old test. @tancheng

I didn't get it. Will you provide a README or not?

n0thingNoob:
> > I am not sure why there is a conflict when pushing the first commit. I merged the latest main into architecture_spec to resolve it and removed the obsolete README that was only used by the old test. @tancheng
> 
> I didn't get it. Will you provide a README or not?

Perhaps not, the branch test is a little bit complicated. Is the README necessary for this test?

n0thingNoob:
> > > > I am not sure why there is a conflict when pushing the first commit. I merged the latest main into architecture_spec to resolve it and removed the obsolete README that was only used by the old test. @tancheng
> > > 
> > > 
> > > I didn't get it. Will you provide a README or not?
> > 
> > 
> > Perhaps not, the branch test is a little bit complicated. Is the README necessary for this test?
> 
> Hi @n0thingNoob, I just meant to have simple explanation maybe about only one tile for two time steps. You don't need explain what happen on all the tiles across all time steps. How does this sound? You can put your README.md back in your test folder, like what @HobbitQia (https://github.com/coredac/dataflow/tree/main/test/visualize) and @ShangkunLi (https://github.com/coredac/dataflow/tree/main/test) did. Sounds good to you?

Sure.

test/code_gen/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+### Core Configuration
+Each core contains:
+- `column`, `row`: Position coordinates in the array
+- `core_id`: Unique identifier for the core
+- `entries`: List of execution contexts for this core
+
+### Instruction Format
+Each `entry` contains:
+- `entry_id` — the (single-instruction) context ID
```
tancheng, line 32: Should we mention for now, each entry can only contain one instruction, and multiple instructions would be supported later?

test/code_gen/README.md
```diff
@@ -0,0 +1,203 @@
...
+  - `timestep` — the per-tile cycle at which the instruction executes
+  - `src_operands` — inputs
+  - `dst_operands` — outputs
+
+Operands encode:
+- `#N` — immediate constant (e.g., `#10`)
+- `$N` — local register (e.g., `$22`)
+- `EAST | WEST | NORTH | SOUTH` — directional ports for inter-tile communication
+
+Each operand also has a `color` field. Colors are primarily meaningful for **directional** operands (to visualize/track routing); for registers and immediates, the color can be ignored by consumers.
```
tancheng, line 44: what do you mean by "**directional** operands"? I thought "color" means different events/kernels/applications.

test/code_gen/README.md
tancheng, line 1: Excellent explanation @n0thingNoob~!

I have a quick question, each instruction has `time_step` (or cycle), how do we automate iterating them as a loop? For example, assuming II = 4, if inst0 has time step 5, it should start execution on cycle 5, then cycle 9, then cycle 13, etc.

So simulator will take the II to perform above iterating? It is actually also okay to not take II, as long as the instructions within on tile can be in-ordered executed (in the order of time step).

test/code_gen/README.md
n0thingNoob: Color is the virtual channel in our original setting of our simulator (because the Cerebras used such a setting at the beginning). It can be used to send different events/kernel/applications. But it is only meaningful when the operand is a direction port because reg does not need this. That's why I said it is for directional operands. For now, we only use the default color.

test/code_gen/README.md
n0thingNoob: But it is a test file README. Do we need to mention the future plan in it?

tancheng: This README.md is not only single test explanation. It is supposed to serve as your codegen design and sample output explanation. You need somewhere to record your design philosophy, limitation, and future plan. Before a proper doc is set up, we have to put everything here.

test/code_gen/README.md
tancheng: The key point is once you complete the entries in the last time step, you will start over the entries in the first time step, right?

And when it knows it could stop simulation?

tancheng:
You are aware of your always failed test, right? https://github.com/coredac/dataflow/actions/runs/17523525173/job/49770978606

And it should be an easy fix for you I think.

test/code_gen/README.md
n0thingNoob: There is a return op, right? Giving a signal after the return to stop the whole system should be easy.I remember we had a feature to know when to stop previously.  I need to discuss with @Jackcuii to make sure.

test/code_gen/README.md
n0thingNoob: ok

n0thingNoob:
> You are aware of your always failed test, right? https://github.com/coredac/dataflow/actions/runs/17523525173/job/49770978606
> 
> And it should be an easy fix for you I think.

Working on it.

test/code_gen/README.md
tancheng: Yes. SGTM.

tancheng:
okay to merge now @n0thingNoob? or anything else you want to fix?

Issue #82, opened 2025-07-17
Agreement on Behavior Representation


Jackcuii:
Here is an agreement on the repr of the PE behavior.
 [https://github.com/n0thingNoob/ZeonicaParser/tree/main/agreement](https://github.com/n0thingNoob/ZeonicaParser/tree/main/agreement)
The Opcode set please refer to the Google Doc

It is a superset of Zeonica's and Neura's function. (room for future)  

Still being edited

Jackcuii:
Hi Sir Tan @tancheng , I still has some questions after the meeting just now. 
I am not quite sure about the predicate embedded data.

> For branch, no need to care predicate-based execution, but do need to care data with predicate embedded data <value, predicate>

Is it means the most traditional dataflow execution? It still seems to need the branches.

tancheng:
> Hi Sir Tan [@tancheng](https://github.com/tancheng) , I still has some questions after the meeting just now. I am not quite sure about the predicate embedded data.
> 
> > For branch, no need to care predicate-based execution, but do need to care data with predicate embedded data <value, predicate>
> 
> Is it means the most traditional dataflow execution? It still seems to need the branches.

Plz take a look at https://github.com/coredac/dataflow/blob/4a6de1792f2dbb51138dadaff3cd46dacb950776/test/neura/ctrl/branch_for.mlir#L61-L102 and try to understand it. Please let me or @ShangkunLi know if you have any questions :-)

Issue #81, opened 2025-07-16
[P1] Assign/model registers in tile


n0thingNoob:
I have a question about how to assign physical registers based on the mapping information from lowered IR.
For example, in the IR we may see %34 = "neura.data_mov"(%29), but the numbers (%29, %34) are just virtual identifiers — they don’t mean we really need 35 physical registers.
In practice, we might still have enough free registers in a tile to accommodate these operations.

So my question is:

How can we determine an appropriate register assignment for each op, ensuring that registers are reused safely where possible and there are no conflicts?

Specifically, what strategy should we follow to assign registers efficiently in each tile, given that the virtual register numbers don’t directly reflect physical register usage?

tancheng:
Discussed offline, we won't specify `R` (register) in instruction for now, and the `mov->add->mov` can actually be represented by single instruction. We will revisit this once the register is correctly model in the `architecture.cc`.

Issue #52, opened 2025-06-19
[P1] Model dataflow accelerator based on specs


tancheng:
The specs can be represented in protobuff/json, which can be used by both compiler and RTL.

YanzhouTang:
> Thanks [@YanzhouTang](https://github.com/YanzhouTang), can you help to check whether the mapper not only respect the size of CGRA, but also respect the supported operations?
> 
> For example, by [default](https://github.com/coredac/dataflow/blob/19f2a170b7f98e49be6344cf5de48bb7ef891666/test/arch_spec/arch_spec_example.yaml#L22), we don't provide ld/st for all the tiles, but override the left-most tiles with ld/st using
> 
> [dataflow/test/arch_spec/arch_spec_example.yaml](https://github.com/coredac/dataflow/blob/19f2a170b7f98e49be6344cf5de48bb7ef891666/test/arch_spec/arch_spec_example.yaml#L58)
> 
> Line 58 in [19f2a17](/coredac/dataflow/commit/19f2a170b7f98e49be6344cf5de48bb7ef891666)
> 
>  operations: ["load"] 
> then the mapper shouldn't map load/store instructions/operations on other tiles:
> 
> [dataflow/include/NeuraDialect/Architecture/Architecture.h](https://github.com/coredac/dataflow/blob/19f2a170b7f98e49be6344cf5de48bb7ef891666/include/NeuraDialect/Architecture/Architecture.h#L110)
> 
> Line 110 in [19f2a17](/coredac/dataflow/commit/19f2a170b7f98e49be6344cf5de48bb7ef891666)
> 
>  bool canSupportOperation(OperationKind operation) const {

I checked the `override` function. After fixing a pointer bug #206 , the `override` can be used correctly now.

For example, if we only activate left-most tiles with ld/st using `override`, our mapper would only map ld/st to left-most tiles.

tancheng:
Thanks @YanzhouTang~!

Can you also help check:
 - What about the multi-cgra, would the cgra mapping only respect the [0, 0] CGRA, if the other cgra config is changed, it won't affect the mapping.
 - What about the tile_override/link_override's [existence](https://github.com/coredac/dataflow/blob/19f2a170b7f98e49be6344cf5de48bb7ef891666/test/arch_spec/arch_spec_example.yaml#L75)? i.e., if a tile or link `existence` field is `false`, would the mapping be affected?

YanzhouTang:
[...]

PR #93, opened 2025-07-26
Prepare Function Unit and Registers for mapping

tancheng:
Prepare the Function Unit and Registers for mapping. A tile could contain registers and functional units
 - hierarchi: `Tile` -> `Register File Cluster` -> `Register File` -> `Register`, https://github.com/coredac/dataflow/issues/81
 - @Jackcuii I have prototyped the Function Unit for you, please honor the arch spec to initialize them, `TODO with @Jackcuii` are left as comments, https://github.com/coredac/dataflow/issues/82 https://github.com/coredac/dataflow/issues/52

TODO:
 - Relax the mapping constraint on the input channel by moving data from the channel to local registers
   - Register might also be modeled in MRRG, https://github.com/coredac/dataflow/issues/81

### DIFF
--- a/include/NeuraDialect/Architecture/Architecture.h
+++ b/include/NeuraDialect/Architecture/Architecture.h
@@ -1,12 +1,14 @@
 #ifndef NEURA_ARCHITECTURE_H
 #define NEURA_ARCHITECTURE_H
 
+#include <cassert>
+#include <map>
+#include <memory>
+#include <optional>
 #include <string>
-#include <vector>
 #include <set>
 #include <unordered_map>
-#include <optional>
-#include <memory>
+#include <vector>
 
 namespace mlir {
 namespace neura {
@@ -15,6 +17,25 @@ namespace neura {
 enum class ResourceKind {
   Tile,
   Link,
+  FunctionUnit,
+  Register,
+  RegisterFile,
+  RegisterFileCluster,
+};
+
+// Enum for function unit resource type.
+enum class FunctionUnitKind {
+  FixedPointAdder,
+  FixedPointMultiplier,
+  CustomizableFunctionUnit,
+};
+
+// Enum for supported operation types.
+enum OperationKind {
+  IAdd = 0,
+  IMul = 1,
+  FAdd = 2,
+  FMul = 3
 };
 
 //===----------------------------------------------------------------------===//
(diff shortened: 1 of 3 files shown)


include/NeuraDialect/Architecture/Architecture.h
```diff
@@ -15,6 +17,25 @@ namespace neura {
 enum class ResourceKind {
   Tile,
   Link,
+  FunctionUnit,
+  Register,
+  RegisterFile,
+  RegisterFileCluster,
+};
```
ShangkunLi, line 24: What's the difference between there 3 kinds of register-related value?

include/NeuraDialect/Architecture/Architecture.h
```diff
@@ -31,7 +52,81 @@ class BasicResource {
...
+
+class FixedPointMultiplier : public FunctionUnit {
+public:
+  FixedPointMultiplier(int id) : FunctionUnit(id) {
+    supported_operations.insert(OperationKind::IMul);
+  }
+  std::string getType() const override { return "fixed_point_multiplier"; }
+  ResourceKind getKind() const override { return ResourceKind::FunctionUnit; }
+};
```
ShangkunLi, line 119: Do we need to add the `cycle` value in the `FunctionUnit` so that we can support multi-cycle function units?

include/NeuraDialect/Architecture/Architecture.h
```diff
@@ -15,6 +17,25 @@ namespace neura {
...
+  RegisterFileCluster,
+};
+
+// Enum for function unit resource type.
+enum class FunctionUnitKind {
+  FixedPointAdder,
+  FixedPointMultiplier,
+  CustomizableFunctionUnit,
+};
+
```
ShangkunLi, line 32: Should we add some resource binding info? Because a function unit can support different operations, and different operations can also be executed in parallel in different function units in a tile if there's no dependency between them.

include/NeuraDialect/Architecture/Architecture.h
tancheng: Sure, it can be done later.

include/NeuraDialect/Architecture/Architecture.h
tancheng: `Tile -> Register File Cluster -> Register File -> Register`, registers are just organized in such hierarchy.

ShangkunLi (approved):

PR #99, opened 2025-07-29
Remove grant_always

ShangkunLi:
In this pr:
- Remove the `grant_always` operations in `--transform-ctrl-to-data-pass`
- After removing the `grant_always`, we can achive `CompiledII = 6 : i32, RecMII = 4 : i32, ResMII = 2 : i32` for `branch_for.mlir` (previous compiledII=7) and `CompiledII = 3 : i32, RecMII = 3 : i32, ResMII = 2 : i32` for `simpleloop.mlir` (previous compiledII=5).

### DIFF
--- a/include/NeuraDialect/NeuraOps.td
+++ b/include/NeuraDialect/NeuraOps.td
@@ -387,10 +387,11 @@ def Neura_LoopControllerOp : Op<NeuraDialect, "loop_controller">{
   }];
 
   let arguments = (ins
-    AnyType:$parent_valid,  // Valid predicate from the parent loop
+    AnyType:$parentValid,  // Valid predicate from the parent loop
     AnyType:$start,         // Start index of the loop
     AnyType:$end,           // End index of the loop
-    AnyType:$step           // Step size for the loop
+    AnyType:$step,           // Step size for the loop
+    StrAttr:$iterationType // Type of the loop iteration (e.g., "increment", "decrement")
   );
 
   let results = (outs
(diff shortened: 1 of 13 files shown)


ShangkunLi:
> Hmm, can this constant binding be done in `FusePatternsPass.cpp`?

I didn't add the constant binding feature in this pr yet. And I creates a single `--fuse-constant-and-grant-once` for the constant binding. Because the ideal flow of fusion is:

Fuse `constant` and `grant_once` -> fuse control flow related operations (e.g., loop control related operations) -> fuse computation related operations (e.g., add-mul, add-add, etc.).

Since I need the ir with fused `constant` and `grant_once` to identify the loop structures, I think we cannot realize it in `FusePatternsPass.cpp`

ShangkunLi:
> > Can you explain a bit on: "Remove the grant_always operations in --transform-ctrl-to-data-pass"?
> > What would it look like if `grant_always` is removed? how is the constant passed in then?
> 
> It looks like "replace" instead of "remove".

We use the `grant_once` for all the live-out constant ops. And for constants used in other blocks, we use `grant_once` + `grant_predicate` to make it valid instead of using `grant_always` like before.

tancheng:
> Since I need the ir with fused constant and grant_once to identify the loop structures, I think we cannot realize it in FusePatternsPass.cpp

I don't get this. Why can't we simply fuse the `constant -> grant_once` pattern? Why do we need to "identify loop structures"?

PR #109, opened 2025-08-13
Add heuristic for no-producer operations

ShangkunLi:
### Background
1. In the existing implementation, we calculate the award based on the producers' location.
2. For operations without any producers, we now select them randomly or solely based on the temporal information.

For example, for `%1 = "neura.grant_once"() <{constant_value = 0 : i64}> : () -> !neura.data<i64, i1>`, it doesn't have any producers, so the candidates for it are chosen randomly or based solely on the time.
```
  Loc: tile#0 @t=0, Award: 8
  Loc: tile#7 @t=0, Award: 8
  Loc: tile#6 @t=0, Award: 8
  Loc: tile#8 @t=0, Award: 8
  Loc: tile#5 @t=0, Award: 8
  Loc: tile#9 @t=0, Award: 8
  Loc: tile#10 @t=0, Award: 8
  Loc: tile#4 @t=0, Award: 8
  Loc: tile#11 @t=0, Award: 8
  Loc: tile#3 @t=0, Award: 8
  Loc: tile#12 @t=0, Award: 8
  Loc: tile#13 @t=0, Award: 8
  Loc: tile#2 @t=0, Award: 8
  Loc: tile#14 @t=0, Award: 8
  Loc: tile#1 @t=0, Award: 8
  Loc: tile#15 @t=0, Award: 8
  Loc: tile#12 @t=1, Award: 7
  Loc: tile#11 @t=1, Award: 7
  Loc: tile#10 @t=1, Award: 7
  Loc: tile#13 @t=1, Award: 7
  Loc: tile#9 @t=1, Award: 7
  Loc: tile#14 @t=1, Award: 7
  Loc: tile#8 @t=1, Award: 7
  Loc: tile#7 @t=1, Award: 7
  Loc: tile#15 @t=1, Award: 7
  Loc: tile#0 @t=1, Award: 7
  Loc: tile#2 @t=1, Award: 7
  Loc: tile#6 @t=1, Award: 7
  Loc: tile#5 @t=1, Award: 7
  Loc: tile#3 @t=1, Award: 7
  Loc: tile#1 @t=1, Award: 7
  Loc: tile#4 @t=1, Award: 7
```

However, these randomly chosen candidate locations may result in sub-optimal mapping results. And if we want to modify the assigned locations for these operations, we need to set a large `max_backtrack_depth` (it takes ~10 hours and is unrealistic).

### New heuristic
1. Calculate the award for no-producer operations based on a `greedy` (`max_locations_to_try=user_defined_num`, equals to 4*4=16 in now implementation, `max_backtrack_depth=0` ) mapping result.
2. Use the number of successfully mapped successor operations as the award. 

And the mapping results show an improvement in both spatial-only and spatial-temporal mapping.

p.s. I also tried an A* search-like heuristic and took tile utilization and routing congestion into consideration, that is, evaluate the award based on both producers and consumers for an operation. But the result is even worse (e.g., `compiled_ii=9` for `branch_for.mlir`), this is because we introduce too many weighted factors, and it is hard to choose appropriate values.

### DIFF
--- a/lib/NeuraDialect/Mapping/HeuristicMapping/HeuristicMapping.cpp
+++ b/lib/NeuraDialect/Mapping/HeuristicMapping/HeuristicMapping.cpp
@@ -160,22 +160,6 @@ HeuristicMapping::NoProducerOpCandidate HeuristicMapping::tryToMapNoProducerOp(
   if (solutions.empty()) {
     llvm::outs() << "[GlobalExploration] No feasible solutions found, "
                     "backtracking\n";
-
-    // snapshots.pop_back();
-    // candidate_history.pop_back();
-    // operation_index_history.pop_back();
-
-    // if (snapshots.empty()) {
-    //   llvm::outs() << "[HeuristicMapping] No more snapshots to restore, "
-    //                   "mapping failed.\n";
-    //   return false;
-    // }
-    // snapshots.back().restore(mapping_state);
-    // candidate_history.back()++;
-
-    // performBacktrack(no_producer_ops, snapshots, candidate_history,
-    //                  operation_index_history, current_op_index,
-    //                  mapping_state);
     return {mapping_state, 0, false, mapping_state.getII()};
   }
 


lib/NeuraDialect/Mapping/HeuristicMapping/HeuristicMapping.cpp
```diff
@@ -78,38 +189,198 @@ bool HeuristicMapping::mapWithBacktrack(
...
+          // Tries to place and route the next operation at each candidate
+          // location.
+          bool mapped = false;
+          for (const auto &next_loc : next_candidates) {
+            if (placeAndRoute(next_op, next_loc, mapping_state)) {
+              mapped = true;
+              mapped_count++;
+              break;
+            }
+          }
```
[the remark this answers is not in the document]
ShangkunLi: We are not performing "real" mapping when we try to find candidate locations for an operation without producers. We just perform a fake mapping to get the score of different candidate locations.

I think mapping the critical path operations first is orthogonal to the heuristic introduced in this pr.

In this pr, I just want to find a reasonable location/candidate for operations without producers, like `constant, grant_once/always <{constant_value=...}>`. In our current implementation, we just randomly pick a location for these operations (because they don't have producers, and we can't differentiate between different locations using existing `calculateAward` function).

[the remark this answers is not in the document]
tancheng: makes sense. Plz try to refactor the code, i.e., wrap the handling of the no-producer ops' trial mapping into helper function, to reflect your reasoning/explanation. The current code is hard to follow, and let's discuss during our meeting.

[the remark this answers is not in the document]
ShangkunLi: Refactor the code it the latest commit~

include/NeuraDialect/Mapping/Mapping.h
```diff
@@ -14,15 +14,28 @@ class Mapping {
...
-  map(std::vector<std::pair<Operation *, int>> &sorted_ops_with_alap_levels,
-      std::set<Operation *> &critical_ops, const Architecture &architecture,
-      MappingState &mapping_state) = 0;
+  virtual bool map(const Architecture &architecture,
+                   MappingState &mapping_state) = 0;
 
   // Gets the name of this strategy
   virtual std::string getName() const = 0;
 
+  void loadDFG(
```
tancheng, line 23: `loadDfg()`

lib/NeuraDialect/Mapping/HeuristicMapping/HeuristicMapping.cpp
```diff
@@ -3,42 +3,295 @@
...
+  // Configures the max number of candidate locations for each no-producer
+  // operation. We set it to a reasonable number based on the architecture size.
+  const int no_producer_candidates_to_try =
+      architecture.getHeight() * architecture.getWidth();
+
+  // Determines the number of candidate locations to try.
+  int candidates_to_try = std::min(static_cast<int>(candidate_locs.size()),
+                                   no_producer_candidates_to_try);
+
+  // Tries full mapping with greedy strategy (All candidate locations andzero
```
tancheng, line 127: `All` -> `all`; `andzero`?

lib/NeuraDialect/Mapping/HeuristicMapping/HeuristicMapping.cpp
```diff
@@ -3,42 +3,295 @@
...
+               << ", fully_mapped="
+               << (best_solution.fully_mapped ? "yes" : "no")
+               << ", estimated_II=" << best_solution.current_ii << "\n";
+
+  // If we have a fully mapped solution, we can finalize the mapping.
+  if (best_solution.fully_mapped) {
+    llvm::outs()
+        << "[HeuristicMapping] Found complete mapping solution through "
+        << "global exploration, finalizing...\n";
+    return true;
```
tancheng, line 200: How do we revert the mapping of the mapped ops (after the no-produce op) during trial?

lib/NeuraDialect/Mapping/HeuristicMapping/HeuristicMapping.cpp
```diff
@@ -3,42 +3,295 @@
...
+
+  if (snapshots.empty()) {
+    llvm::outs() << "[HeuristicMapping] No more snapshots to restore, "
+                 << "mapping failed.\n";
+    return false; // No more snapshots to restore, mapping failed.
+  }
+
+  // Because we have already thoroughly iterated candidates of no-producer
+  // operations. If the current operation is after a no-producer operation, we
+  // need backtrack 2 depths.
```
tancheng, line 229: Why? And backtrack 2 depth/layer means a no-producer op is also backtracked?

ShangkunLi: I see the bug here. If `current_op` fails, now we will backtrack to the mapping state before mapping `no_producer_op1`; it still is in an infinite loop.

The correct solution is to backtrack until it is not a `no_producer_op`, because only in this case can we result in an effective backtrack.

tancheng: Hi @ShangkunLi, could you please add an example as comments above the function, like:
```
<no_producer_op0, no_producer_op1, no_producer_op2, op3, op4>

          |              |                  |        |    |

      | loc_0 |      | loc_3 | ...
      | loc_1 |      | loc_4 | ...
      | loc_2 |      | loc_5 | ...

```
explain what happen for different cases:
 - op4 failed
 - op3 failed
 - no_producer_op2 failed

tancheng: I am a bit hesitate to check in this feature... Such looking ahead is too complicated for our code structure, and error-prone... I believe simply assigning initial loc for no-producer-op might be able to achieve similar benefit...

lib/NeuraDialect/Mapping/HeuristicMapping/HeuristicMapping.cpp
```diff
@@ -225,25 +230,51 @@ bool HeuristicMapping::performBacktrack(
...
+
+  //   |         |                 |        |    |
+
+  //         | loc_0 |         | loc_3 | ...
+  //         | loc_1 |         | loc_4 | ...
+  //         | loc_2 |         | loc_5 | ...
+
+  // 1. If we fail to map op4, we can backtrack to op3 using this
+  // performBacktrack function.
+  // 2. If we fail to map op3, we need to backtrack to op0 to avoid getting
```
tancheng, line 232: Why we would have `op0` before `no_producer_op1`? Shouldn't ops be ordered in topological ordering?

tancheng:
Let's check in the PR of "the critical path mapping first". Thanks~!

PR #115, opened 2025-08-19
Allocate registers for data_mov on the same tile

ShangkunLi:
In this pr:
- Add special logic for `data_mov` or `ctrl_mov` on the same tile
- Specifically, we just allocate registers for such data passed through `data_mov` or `ctrl_mov`

### DIFF
--- a/lib/NeuraDialect/Mapping/mapping_util.cpp
+++ b/lib/NeuraDialect/Mapping/mapping_util.cpp
@@ -427,32 +427,34 @@ bool mlir::neura::tryRouteDataMove(Operation *mov_op, MappingLoc src_loc,
 
     // Defines the time window for the register allocation.
     int start_time = src_loc.time_step;
-    int end_time = dst_loc.time_step;
+    int exclusive_end_time = dst_loc.time_step;
 
-    // For backward moves, we need to adjust the end time
+    // For backward moves, we need to adjust the end time.
     if (is_backward_move) {
-      end_time += state.getII();
+      exclusive_end_time += state.getII();
     }
 
-    // Finds a register that is available during start_time to end_time
-    Register *reg = getAvailableRegister(state, tile, start_time, end_time);
+    // Finds a register that is available during start_time to
+    // exclusive_end_time.
+    Register *reg =
+        getAvailableRegister(state, tile, start_time, exclusive_end_time);
     if (!reg) {
       llvm::errs() << "[tryRouteDataMove] No available register found for "
                    << "tile: " << tile->getId()
                    << " from time step: " << start_time
-                   << " to time step: " << end_time << "\n";
+                   << " to time step: " << exclusive_end_time << "\n";
       return false;
     }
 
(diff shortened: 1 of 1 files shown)


lib/NeuraDialect/Mapping/mapping_util.cpp
```diff
@@ -417,6 +417,42 @@ bool mlir::neura::tryRouteDataMove(Operation *mov_op, MappingLoc src_loc,
...
+    // allocate registers for data movement.
+    Tile *tile = dyn_cast<Tile>(src_loc.resource);
+    if (!tile) {
+      llvm::errs() << "[tryRouteDataMove] Source is not a tile\n";
+      return false;
+    }
+
+    // Defines the time window for the register allocation.
+    int start_time = src_loc.time_step;
+    int end_time = dst_loc.time_step;
```
tancheng, line 430: Shouldn't we use:
```
int exclusive_end_time = dst_loc.time_step - 1;
```

lib/NeuraDialect/Mapping/mapping_util.cpp
```diff
@@ -417,6 +417,42 @@ bool mlir::neura::tryRouteDataMove(Operation *mov_op, MappingLoc src_loc,
...
+
+    // For backward moves, we need to adjust the end time
+    if (is_backward_move) {
+      end_time += state.getII();
+    }
+
+    // Finds a register that is available during start_time to end_time
+    Register *reg = getAvailableRegister(state, tile, start_time, end_time);
```
tancheng, line 438:
```
Register *reg = getAvailableRegister(state, tile, start_time, exclusive_end_time);
```

lib/NeuraDialect/Mapping/mapping_util.cpp
```diff
@@ -417,6 +417,42 @@ bool mlir::neura::tryRouteDataMove(Operation *mov_op, MappingLoc src_loc,
...
+    if (!reg) {
+      llvm::errs() << "[tryRouteDataMove] No available register found for "
+                   << "tile: " << tile->getId()
+                   << " from time step: " << start_time
+                   << " to time step: " << end_time << "\n";
+      return false;
+    }
+
+    // Adds the register locations to the path.
+    for (int t = start_time; t < end_time; ++t) {
```
tancheng, line 448:
```
for (int t = start_time; t <= exclusive_end_time; ++t) {
```

lib/NeuraDialect/Mapping/mapping_util.cpp
ShangkunLi: Yes

tancheng: why, sounds natural to me.. it indicates "end" but not inclusive itself, and the func `getAvailableRegister`'s signature also has `exclusive` in its naming.

tancheng (approved):

PR #126, opened 2025-09-16
Fuse the RHS constant into corresponding op

ShangkunLi:
In this pr:
- Provide a template to fuse the RHS constant into its operation
- Separate the optimization pass as HWAgnostic optimization (e.g., `fold-constant`) and HWSpecific optimization (e.g., `fuse-loop-control`)
- I have removed some predicate attributes in the `NeuraOps.td` because we need to enable the `$rhs` as an optional operand. And multiple optional operands are not allowed in mlir.

After applying the `fold-constant`, the number of ops can be greatly reduced.

For dataflow `branch_for.mlir`, 40 lines -> 25 lines.
For dataflow `simple_loop.mlir`, 53 lines -> 29 lines.


lib/NeuraDialect/Transforms/Optimizations/HWAgnosticOpt/CMakeLists.txt
tancheng, line 1: HWAgnosticOpt -> HwAgnosticOpt

lib/NeuraDialect/Transforms/Optimizations/HWAgnosticOpt/FoldConstantPass.cpp
```diff
@@ -0,0 +1,252 @@
...
+  return constant_op->getAttr("value");
+}
+
+void addConstantAttribute(Operation *op, StringRef attr_name,
+                          Attribute const_value) {
+  op->setAttr(attr_name, const_value);
+}
+
+// A template pattern to fuse binary operations with a constant on the
+// right-hand side.
```
tancheng, line 99: right-hand side operand.

lib/NeuraDialect/Transforms/Optimizations/HWAgnosticOpt/FoldConstantPass.cpp
```diff
@@ -0,0 +1,252 @@
...
+  using OpRewritePattern<OpType>::OpRewritePattern;
+
+  virtual Operation *
+  createOpWithFusedRhsConstant(OpType op, Attribute rhs_const_value,
+                               PatternRewriter &rewriter) const = 0;
+
+  LogicalResult matchAndRewrite(OpType op,
+                                PatternRewriter &rewriter) const override {
+    Value lhs = op.getLhs();
+    Value rhs = op.getRhs();
```
tancheng, line 111: Shouldn't we have?
```
if op.getNumOperands() < 2:
  return failure();
```

lib/NeuraDialect/Transforms/Optimizations/HWSpecificOpt/CMakeLists.txt
tancheng, line 1: HwSpecificOpt

lib/NeuraDialect/Transforms/PromoteFuncArgToConstPass.cpp
tancheng, line 1: What if a func takes a variable rather than constant as argument?

lib/NeuraDialect/Transforms/Optimizations/HWAgnosticOpt/FoldConstantPass.cpp
tancheng: And is that true all of our ops have getLhs() and getRhs()? We don't have op has single operand, or more than 2 operands?

test/neura/ctrl/branch_for.mlir
```diff
@@ -96,251 +102,181 @@ func.func @loop_test() -> f32 {
...
-// MAPPING-NEXT:     %47 = "neura.data_mov"(%45) {mapping_locs = [{id = 34 : i32, resource = "link", time_step = 4 : i32}, {id = 56 : i32, resource = "register", time_step = 5 : i32}]} : (!neura.data<i1, i1>) -> !neura.data<i1, i1>
-// MAPPING-NEXT:     %48 = neura.grant_predicate %46, %47 {mapping_locs = [{id = 14 : i32, resource = "tile", time_step = 6 : i32, x = 2 : i32, y = 3 : i32}]} : !neura.data<f32, i1>, !neura.data<i1, i1> -> !neura.data<f32, i1>
-// MAPPING-NEXT:     %49 = "neura.data_mov"(%48) {mapping_locs = [{id = 43 : i32, resource = "link", time_step = 6 : i32}]} : (!neura.data<f32, i1>) -> !neura.data<f32, i1>
-// MAPPING-NEXT:     "neura.return"(%49) {mapping_locs = [{id = 13 : i32, resource = "tile", time_step = 7 : i32, x = 1 : i32, y = 3 : i32}]} : (!neura.data<f32, i1>) -> ()
+// MAPPING:       func.func @loop_test() -> f32 attributes {accelerator = "neura", mapping_info = {compiled_ii = 5 : i32, mapping_mode = "spatial-temporal", mapping_strategy = "heuristic", rec_mii = 4 : i32, res_mii = 1 : i32, x_tiles = 4 : i32, y_tiles = 4 : i32}} {
```
tancheng, line 211: lol, still 5? :-(

test/neura/interpreter/basic_operation/add.mlir
```diff
@@ -46,15 +46,17 @@ func.func @test_add_zero() -> f32 {
 
 // RUN: neura-interpreter %s | FileCheck %s
 
+// Remove Test 5 because we plan to remove the predicate attribute in
```
tancheng, line 49: // TODO: Remove...

lib/NeuraDialect/Transforms/Optimizations/HWAgnosticOpt/FoldConstantPass.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+    if (isOriginConstantOp(lhs)) {
+      assert(false && "LHS constant folding not implemented yet.");
+      return failure();
+    }
+
+    if (!rhs || !isOriginConstantOp(rhs)) {
+      return failure();
+    }
```
ShangkunLi, line 120: > Shouldn't we have?
> 
> ```
> if op.getNumOperands() < 2:
>   return failure();
> ```

We guarantee the operation has a valid `$rhs` here.

lib/NeuraDialect/Transforms/Optimizations/HWAgnosticOpt/FoldConstantPass.cpp
ShangkunLi: All operands of the binary operations in our definition are defined as `$lhs` and `$rhs`.  So when we face an operation like
```
def Neura_FAddFAddOp : Op<NeuraDialect, "fadd_fadd"> {
  let summary = "Fused fadd(fadd(a, b), c)";
  let arguments = (ins AnyType:$a, AnyType:$b, AnyType:$c, Optional<AnyType>:$predicate);
  let results = (outs AnyType:$result);
  // let assemblyFormat = "$a `,` $b `,` $c `,` $predicate attr-dict `:` type($result)";
  let traits = [SameOperandsAndResultElementType];
}
```

We cannot access its operands through `getLhs()` and `getRhs()`

lib/NeuraDialect/Transforms/PromoteFuncArgToConstPass.cpp
ShangkunLi: Hmmm, I promote function args to constants because I view it as value propagation like c/c++ code.

I haven't considered how to support point handling for now.....

test/neura/ctrl/branch_for.mlir
tancheng: Soudns good. We can refactor the mapping stuff, to make it easily to plug in different mapping algorithms, then check in your code later.

lib/NeuraDialect/Transforms/PromoteFuncArgToConstPass.cpp
ShangkunLi: It has been used many times in our `canonicalize-live-in` pass. I extract it as a single pass for further function arguments fusing.
https://github.com/coredac/dataflow/blob/7f845d6a5b9dd3310c237dba5db9934b032f14f4/lib/NeuraDialect/Transforms/CanonicalizeLiveInPass.cpp#L20

For runtime values, you mean a reference? I haven't considered that yet...

This pass creates a constant op for each function argument, and removes the function arguments from the `func.func`. If we do not do this, we cannot handle mapping properly. In fact, we can fuse these arguments into corresponding ops if we view them as value propagation.

tancheng: I didn't see where `--promote-func-arg-to-const` is used in our tests? or  you extract it in this PR, but will remove the old code and leverage the flag in next PR?

ShangkunLi: Yes, will do it in the next PR

PR #135, opened 2025-09-27
Add common pattern fusion

HobbitQia:
Fuse `gep` + `load`/`store`, `mul` + `add` in `FusePatternPass` which I observed them in most kernels and provide a C++ test case to verify the correctness.

For now, the fusion should be written manually, I am working on searching on common patterns based on frequency and fusing different pattern into the same op but with different attributes to differentiate each other. I will finish this these days.

### DIFF
--- a/lib/NeuraDialect/Transforms/FusePatternPass.cpp
+++ b/lib/NeuraDialect/Transforms/FusePatternPass.cpp
@@ -134,7 +134,7 @@ struct FuseGepLoadPattern : public OpRewritePattern<neura::LoadOp> {
     }
     
     auto fused = rewriter.create<neura::LoadIndexedOp>(
-        loc, type, gep_op.getBase(), indexValues, load.getPredicate());
+        loc, type, gep_op.getBase(), indexValues);
 
     rewriter.replaceOp(load, fused.getResult());
     rewriter.eraseOp(gep_op);
@@ -166,7 +166,7 @@ struct FuseGepStorePattern : public OpRewritePattern<neura::StoreOp> {
     }
     
     rewriter.create<neura::StoreIndexedOp>(
-        loc, store.getValue(), gep_op.getBase(), indexValues, store.getPredicate());
+        loc, store.getValue(), gep_op.getBase(), indexValues);
 
     rewriter.eraseOp(store);
     rewriter.eraseOp(gep_op);


lib/NeuraDialect/Transforms/FusePatternPass.cpp
```diff
@@ -99,6 +109,119 @@ struct FuseFMulFAddPattern : public OpRewritePattern<neura::FAddOp> {
   }
 };
 
+struct FuseGEPLoadPattern : public OpRewritePattern<neura::LoadOp> {
```
tancheng, line 112: `FuseGEPLoadPattern` -> `FuseGepLoadPattern`

lib/NeuraDialect/Transforms/FusePatternPass.cpp
```diff
@@ -99,6 +109,119 @@ struct FuseFMulFAddPattern : public OpRewritePattern<neura::FAddOp> {
...
+    auto fused = rewriter.create<neura::LoadIndexedOp>(
+        loc, type, gep_op.getBase(), indexValues, load.getPredicate());
+
+    rewriter.replaceOp(load, fused.getResult());
+    rewriter.eraseOp(gep_op);
+    return success();
+  }
+};
+
+struct FuseGEPStorePattern : public OpRewritePattern<neura::StoreOp> {
```
tancheng, line 145: `FuseGEPStorePattern` -> `FuseGepStorePattern`

test/neura/fusion/kernel.cpp
```diff
@@ -0,0 +1,26 @@
+#include <stdio.h>
+#include <stdlib.h>
+
+#define NTAPS 1024
+
+int A[NTAPS][NTAPS];
+int s[NTAPS];
+int q[NTAPS];
+int p[NTAPS];
+int r[NTAPS];
```
tancheng, line 10: Meaningless kernel with meaningless variables? Just to verify the fusion functionality, right?

test/neura/fusion/test.mlir
```diff
@@ -0,0 +1,19 @@
...
+# RUN: mlir-neura-opt --assign-accelerator \
+# RUN:           --lower-llvm-to-neura \
+# RUN:           --canonicalize-live-in \
+# RUN:           --leverage-predicated-value \
+# RUN:           --fold-constant \
+# RUN:           --transform-ctrl-to-data-flow \
+# RUN:           --fold-constant \
+# RUN:           --fuse-pattern \
+# RUN:           --view-op-graph \
+# RUN:           --insert-data-mov kernel.mlir | FileCheck %s --check-prefix=CHECK-FUSED
```
tancheng, line 12: Can you please copy the generated DFG.png in the PR's description (before and after), so we can visualize the fusion?

tancheng:
> fusing different pattern into the same op but with different attributes to differentiate each other.

This sounds a good idea, which is different from creating every `neura.new_op` for each fused op. However, we then need to carefully put sufficient/many attributes to "differentiate each other". Anyways, the design about those attributes can be first prototyped in a simple way and then improved gradually.

test/neura/fusion/test.mlir
tancheng, line 12: Can we also try to map it? Trying to see what might be the blocker for mapping arbitrary kernel.

lib/NeuraDialect/Transforms/FusePatternPass.cpp
HobbitQia: fixed

test/neura/fusion/test.mlir
HobbitQia: yes! now it can be successfully mapped.

lib/NeuraDialect/Transforms/FusePatternPass.cpp
tancheng: `FuseGEPStorePattern` is still not fixed.

test/neura/fusion/test.mlir
```diff
@@ -0,0 +1,24 @@
...
+# RUN: mlir-neura-opt --map-to-accelerator="mapping-strategy=heuristic backtrack-config=customized" %t-kernel_dataflow.mlir | FileCheck %s --check-prefix=CHECK-MAPPING
+
+# CHECK-FUSED: func.func
+# CHECK-FUSED: accelerator = "neura"
+# CHECK-FUSED: %102 = neura.load_indexed %100[%101 : !neura.data<i64, i1>] !neura.data<!llvm.ptr, i1> : !neura.data<i32, i1>
+# CHECK-FUSED: %33 = "neura.mul_add"(%30, %31, %32) : (i32, i32, i32) -> i32
+# CHECK-FUSED: %42 = "neura.mul_add"(%39, %40, %41) : (i32, i32, i32) -> i32
+
+# CHECK-MAPPING: mapping_info 
+# CHECK-MAPPING: mapping_mode = "spatial-temporal", mapping_strategy = "heuristic", rec_mii = 9 : i32, res_mii = 5 : i32, x_tiles = 4 : i32, y_tiles = 4 : i32
```
tancheng, line 23: what is the `mapped_ii`?

test/neura/fusion/test.mlir
```diff
@@ -0,0 +1,24 @@
+# RUN: clang++ -S -emit-llvm -O3 -fno-unroll-loops -fno-vectorize -o %t-kernel.ll kernel.cpp
```
tancheng, line 1: `-fno-unroll-loops -fno-vectorize` these two really matter? It would be vectorized or unrolled if you don't add them?

test/neura/fusion/test.mlir
HobbitQia: I deliberately remove it from the checking since in the future the mapping algorithm may be modified so the mapped II will also be improved and we need to modify this test file. What do u think of this?

test/neura/fusion/test.mlir
tancheng: It is only vectorized when I add `-f-vectorize` on my side. Anyways, LGTM. Thanks.

PR #138, opened 2025-09-30
Support more test cases for mapping

ShangkunLi:
In this pr, we make the following changes to make it support more mapping cases:

- Add `neura.shl` and `neura.rem` operations
- Add more constant folding patterns
- Make the `neura.alloca` supporting lowering from `memref.alloca`
- Increase the `maxII` to 20
- Increase the number of registers in each tile to 32

### DIFF
--- a/lib/Conversion/MemRefToNeura/MemRefToNeuraPass.cpp
+++ b/lib/Conversion/MemRefToNeura/MemRefToNeuraPass.cpp
@@ -8,6 +8,7 @@
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/Pass/Pass.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "llvm/Support/LogicalResult.h"
 
 using namespace mlir;
 using namespace mlir::neura;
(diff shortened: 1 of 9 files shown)


lib/Conversion/MemRefToNeura/MemRefToNeuraPass.cpp
```diff
@@ -46,6 +47,30 @@ struct MemRefStoreLowering : public OpRewritePattern<memref::StoreOp> {
   }
 };
 
+struct MemRefAllocaToNeuraAlloca : public OpRewritePattern<memref::AllocaOp> {
```
tancheng, line 50: Can we have a design or plan about how to proper represent alloc? Can you help provide a simple memref alloc example, so I can recall what it is used for, then we can think about its functionality in neura land?

tancheng: Thanks @ShangkunLi. Can you also provide the corresponding C++ if there is? or whatever high level this alloc is lowered from?

Dataflow architecture can also have memory management (i take my previous statement back:-)). The design should be about whether we want it is preconfigured (statically) managed or dynamically managed in our scrachpad memory. I think dynamically managed is also possible, as we have "controller" in our architecture, though the first time execution of alloc would take more cycles/latency, while the later execution (due to repeated execution in dataflow) would just take 1 cycle (nothing to do actually).

tancheng: Does the `%alloca = memref.alloca() : memref<256xi64>` represent the `queue[N]`? Why the queue is first accessed w/o being written? Is the C++ correct?

ShangkunLi: Yes, the alloca operation represents `queue[N]`.

Oh, because I removed the `queue[0] = source;` for test reasons...

tancheng: okay. The alloc can be supported by our HW, plz file an issue in VectorCGRA repo. Thanks~!

PR #146, opened 2025-10-08
Support Customized DFG Sorting Strategy

ShangkunLi:
Due to the time limit, I just enabled the customized sorting strategy for steering-based dataflow IR mapping.

Maybe we can properly design the mixed sorting later. For now, I just use topologically sorted ops for evaluation.

### DIFF
--- a/lib/NeuraDialect/Transforms/MapToAcceleratorPass.cpp
+++ b/lib/NeuraDialect/Transforms/MapToAcceleratorPass.cpp
@@ -40,6 +40,11 @@ struct MapToAcceleratorPass
   MapToAcceleratorPass() = default;
   MapToAcceleratorPass(const MapToAcceleratorPass &pass)
       : PassWrapper<MapToAcceleratorPass, OperationPass<ModuleOp>>(pass) {}
+  Option<std::string> sortStrategy{
+      *this, "sort-strategy",
+      llvm::cl::desc("Strategy for sorting operations before mapping. "
+                     "Options: topological, mixed (default)."),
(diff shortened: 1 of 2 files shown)


lib/NeuraDialect/Transforms/MapToAcceleratorPass.cpp
```diff
@@ -206,30 +212,36 @@ struct MapToAcceleratorPass
...
-          getOpsInAlapLevels(topologically_sorted_ops, critical_ops);
-      for (int level = 0; level < static_cast<int>(level_buckets.size());
-           ++level) {
-        llvm::outs() << "[MapToAcceleratorPass] ALAP Bucket Level " << level
-                     << ": " << level_buckets[level].size() << " ops\n";
-        for (Operation *op : level_buckets[level]) {
-          llvm::outs() << "  " << *op << "\n";
+
+      std::vector<std::pair<Operation *, int>> sorted_ops_with_levels;
+      if (sort_strategy_stringRef == "topological") {
```
tancheng, line 217: Variable name `sort_strategy_stringRef` -> `sort_strategy_string_ref`. Corresponding to the `attribute.h` (the issue you filed), we should have a enum to capsulate all these different choices, instead of directly check string equal.

tancheng, line 217: Add comment above mentioning only two sorting algorithms for now, and by default is `mixed`.

tancheng (approved):

Issue #154, opened 2025-10-14
[P0] Codegen pass mishandles promoted func args (neura.constant {value="%arg*"})


n0thingNoob:
# Summary

When function arguments are promoted to constants (via --promote-func-arg-to-const), the IR contains neura.constant ops whose value is a string like "%arg0". The current codegen incorrectly treats these as numeric immediates (e.g., #2) and/or yields an empty src in the generated YAML, causing downstream consumers to “not recognize” the operand.

**Expected:** codegen should preserve %arg* as symbolic operands and respect their mapping info.
**Actual:** %arg* is collapsed to a numeric immediate or results in an empty src.

# Problem Description

When writing the FFT kernel like this:

void fft_cgra_kernel(float *data, const float *tw, long long N);

Compile/transform with the pass pipeline that includes --promote-func-arg-to-const.

**Observe IR (example):**

%0 = "neura.constant"() <{value = "%arg0"}> {mapping_locs = [{id = 2 : i32, resource = "tile", time_step = 12 : i32, x = 2 : i32, y = 0 : i32}]} : () -> !neura.data<!llvm.ptr, i1>
%1 = "neura.constant"() <{value = "%arg1"}> {mapping_locs = [{id = 14 : i32, resource = "tile", time_step = 12 : i32, x = 2 : i32, y = 3 : i32}]} : () -> !neura.data<!llvm.ptr, i1>
%2 = "neura.constant"() <{value = "%arg2"}> {mapping_locs = [{id = 10 : i32, resource = "tile", time_step = 12 : i32, x = 2 : i32, y = 2 : i32}]} : () -> !neura.data<i64, i1>

Run the codegen pass.

Inspect the emitted ASM/YAML: %arg* is turned into an immediate like #2 or the YAML has an empty src operand. (Fixed)

# Current Solution
 We currently directly output %arg* for both YAML and ASM files like this:
**YAML**
                - opcode: "CONSTANT"
                  src_operands:
                    - operand: "%arg1"
                      color: "RED"
                  dst_operands:
                    - operand: "NORTH"
                      color: "RED"
**ASM**
PE(3,1):
{
  CONSTANT, [%arg1, RED] -> [NORTH, RED]
} (t=3)

However, when implementing this in the simulator, we might need to further discuss this issue.

Issue #152, opened 2025-10-13
[P0] Potential redundent oprations optimization


YanzhouTang:
I've been testing c2llvm2mlir testcase recently, and I found that there are some redundent oprations after convering to dataflow format mlir that can be potentially optimized.

Take [test/c2llvm2mlir/simple_loop](../tree/main/test/c2llvm2mlir/simple_loop) for example:

The original mlir is:

```
  llvm.func local_unnamed_addr @_Z6kernelPfS_S_(%arg0: !llvm.ptr {llvm.nocapture, llvm.noundef, llvm.readonly}, %arg1: !llvm.ptr {llvm.nocapture, llvm.noundef}, %arg2: !llvm.ptr {llvm.nocapture, llvm.noundef, llvm.readonly}) attributes {memory_effects = #llvm.memory_effects<other = none, argMem = readwrite, inaccessibleMem = none>, no_unwind, passthrough = ["mustprogress", "nofree", "norecurse", "nosync", ["uwtable", "2"], ["min-legal-vector-width", "0"], ["no-trapping-math", "true"], ["stack-protector-buffer-size", "8"], ["target-cpu", "x86-64"]], target_cpu = "x86-64", target_features = #llvm.target_features<["+cmov", "+cx8", "+fxsr", "+mmx", "+sse", "+sse2", "+x87"]>, tune_cpu = "generic"} {
    %0 = llvm.mlir.constant(0 : i64) : i64
    %1 = llvm.mlir.constant(1 : i64) : i64
    %2 = llvm.mlir.constant(32 : i64) : i64
    %3 = llvm.load %arg1 {alignment = 4 : i64, tbaa = [#tbaa_tag]} : !llvm.ptr -> f32
    llvm.br ^bb1(%0, %3 : i64, f32)
  ^bb1(%4: i64, %5: f32):  // 2 preds: ^bb0, ^bb1
    %6 = llvm.getelementptr inbounds %arg0[%4] : (!llvm.ptr, i64) -> !llvm.ptr, f32
    %7 = llvm.load %6 {alignment = 4 : i64, tbaa = [#tbaa_tag]} : !llvm.ptr -> f32
    %8 = llvm.getelementptr inbounds %arg2[%4] : (!llvm.ptr, i64) -> !llvm.ptr, f32
    %9 = llvm.load %8 {alignment = 4 : i64, tbaa = [#tbaa_tag]} : !llvm.ptr -> f32
    %10 = llvm.fmul %7, %9 : f32
[...]

PR #155, opened 2025-10-14
Explictly use promote-func-arg-to-constant pass

ShangkunLi:
In this PR, we enable the explicit `promote-func-arg-to-constant` pass in tests.

Related issues:
 - https://github.com/coredac/dataflow/issues/154
 - https://github.com/coredac/dataflow/issues/152

### DIFF
--- a/lib/NeuraDialect/NeuraPasses.cpp
+++ b/lib/NeuraDialect/NeuraPasses.cpp
@@ -27,6 +27,7 @@ void mlir::neura::registerNeuraConversionPassPipeline() {
 
         pm.addPass(mlir::neura::createCanonicalizeCastPass());
         pm.addPass(mlir::neura::createPromoteFuncArgToConstPass());
+        pm.addPass(mlir::neura::createFoldConstantPass());
         pm.addPass(mlir::neura::createCanonicalizeLiveInPass());
         pm.addPass(mlir::neura::createLeveragePredicatedValuePass());
         pm.addPass(mlir::createPrintOpGraphPass(os));
--- a/lib/NeuraDialect/Transforms/Optimizations/HwAgnosticOpt/FoldConstantPass.cpp
+++ b/lib/NeuraDialect/Transforms/Optimizations/HwAgnosticOpt/FoldConstantPass.cpp
@@ -111,7 +111,7 @@ struct FuseRhsConstantPattern : public OpRewritePattern<OpType> {
     Value rhs = op.getRhs();
 
     if (isOriginConstantOp(lhs)) {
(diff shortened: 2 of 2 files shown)


lib/NeuraDialect/NeuraPasses.cpp
```diff
@@ -26,6 +26,7 @@ void mlir::neura::registerNeuraConversionPassPipeline() {
         pm.addPass(mlir::createPrintOpGraphPass(os));
 
         pm.addPass(mlir::neura::createCanonicalizeCastPass());
+        pm.addPass(mlir::neura::createPromoteFuncArgToConstPass());
```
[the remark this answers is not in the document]
ShangkunLi: Good idea~, I added it in the latest commit.

lib/NeuraDialect/Transforms/Optimizations/HwAgnosticOpt/FoldConstantPass.cpp
```diff
@@ -111,7 +111,7 @@ struct FuseRhsConstantPattern : public OpRewritePattern<OpType> {
     Value rhs = op.getRhs();
 
     if (isOriginConstantOp(lhs)) {
-      assert(false && "LHS constant folding not implemented yet.");
+      llvm::errs() << "LHS constant folding not supported yet.\n";
```
tancheng: Why relax this constraint?

ShangkunLi: Because we just want to avoid fusing lsh into operation. The previous assertion means we do not allow such operations (`add(const, variable)`) to exist.

tancheng: Got it, thanks. We actually can swap them and then fusion (can happen within this pass instead of two passes) for add/mul (but not div/shift, etc).

tancheng: @YanzhouTang plz include this folding in your PR (unless @ShangkunLi is already doing so?). Let me know if you encounter any blocker when folding the arg_const. Note that we prefer series of tiny PRs (easy to review) rather than tons of code in single PR.

ShangkunLi: I can do it soon~ Just let yanzhou focus on the `arg_const` folding

Issue #145, opened 2025-10-08
[P0] Add attributes header file to manage attributes in neura dialects


ShangkunLi:
We need to create a header file for many attributes (e.g., `dataflow_mode`, `mapping_mode`, `mapping_strategy`, etc.) in our dialect.

YanzhouTang:
Hi Shangkun, is this new feature supports float type now? I tested it in `test/c2llvm2mlir/simple_loop` but failed. Here below is my `test-mlir` in `test/c2llvm2mlir/simple_loop` 

```
// Compiles the original kernel.
// RUN: clang++ kernel.cpp -o %t-kernel.out

// Compiles the original kernel to mlir, then lower back to llvm, eventually binary.
// RUN: clang++ -S -emit-llvm -o %t-kernel.ll kernel.cpp
// RUN: mlir-translate --import-llvm %t-kernel.ll -o %t-kernel.mlir
// RUN: mlir-opt %t-kernel.mlir | mlir-translate -mlir-to-llvmir -o %t-kernel_back.ll
// RUN: llc %t-kernel_back.ll -relocation-model=pic -filetype=obj -o %t-kernel_back.o
// RUN: clang++ %t-kernel_back.o -o %t-kernel_back.out

// RUN: %t-kernel.out > %t-dumped_output.txt
// RUN: %t-kernel_back.out >> %t-dumped_output.txt
// RUN: FileCheck %s < %t-dumped_output.txt

// Verifies the output values are the same for the original and re-compiled kernel.
// CHECK: output: [[OUTPUT:[0-9]+\.[0-9]+]]
// CHECK: output: [[OUTPUT]]

// Test LLVM to NEURA lowering
// RUN: clang++ -S -emit-llvm -ffp-contract=off kernel.cpp -o %t-kernel.ll
// RUN: mlir-translate --import-llvm %t-kernel.ll -o %t-kernel.mlir

// RUN: mlir-neura-opt --assign-accelerator \
// RUN:   --lower-llvm-to-neura \
// RUN:   --canonicalize-live-in \
// RUN:   --leverage-predicated-value \
// RUN:   --transform-ctrl-to-data-flow \
// RUN:   --fold-constant \
// RUN:   --transform-to-steer-control \
// RUN:   --remove-predicated-type \
// RUN:   --insert-data-mov %t-kernel.mlir -o %t-kernel-neura.mlir
[...]

PR #157, opened 2025-10-16
Add promoted const_args to --fold-constant PASS

YanzhouTang:
As we discussed in Issue #152 , after args being promoted to constant, they are not folded in the next `--fold-constant` PASS. 

This PR add this feature based on [simple_loop](https://github.com/coredac/dataflow/tree/main/test/c2llvm2mlir/simple_loop) case.

1. `neura.gep`: its left operand represents base_pointer, we fold this operand if it is a constant.
2. `neura.store`: its right operand represents destination address, we fold this operand if it is a constant.

After leveraging `--promote-func-arg-to-const` and `--fold-constant`, the final operations are reduced to 20, and II = 5 (equals to RecII = 5)

### DIFF
--- a/lib/NeuraDialect/Transforms/Optimizations/HwSpecificOpt/FuseLoopControlPass.cpp
+++ b/lib/NeuraDialect/Transforms/Optimizations/HwSpecificOpt/FuseLoopControlPass.cpp
@@ -106,8 +106,8 @@ private:
 // Finds the constant attribute for a value.
 Attribute findConstantAttribute(Operation *op) {
   // Checks if the operation has a constant attribute.
-  if (op && op->hasAttr("rhs_const_value")) {
-    return op->getAttr("rhs_const_value");
+  if (op && op->hasAttr("rhs_value")) {
+    return op->getAttr("rhs_value");
   }
 
   // If the value is already a constant, return it.
--- a/test/controflow_fuse/simple_loop/simple_loop.mlir
+++ b/test/controflow_fuse/simple_loop/simple_loop.mlir
@@ -196,8 +196,8 @@ module attributes {} {
 // FUSE-NEXT:     %7 = neura.grant_predicate %5, %valid : !neura.data<memref<?xi32>, i1>, !neura.data<i1, i1> -> !neura.data<memref<?xi32>, i1>
 // FUSE-NEXT:     %8 = neura.grant_predicate %3, %valid : !neura.data<memref<?xi32>, i1>, !neura.data<i1, i1> -> !neura.data<memref<?xi32>, i1>
 // FUSE-NEXT:     %9 = neura.load_indexed %7[%nextindex : !neura.data<i64, i1>] !neura.data<memref<?xi32>, i1> : !neura.data<i32, i1>
-// FUSE-NEXT:     %10 = "neura.mul"(%9) {rhs_const_value = 2 : i32} : (!neura.data<i32, i1>) -> !neura.data<i32, i1>
-// FUSE-NEXT:     %11 = "neura.add"(%10) {rhs_const_value = 1 : i32} : (!neura.data<i32, i1>) -> !neura.data<i32, i1>
+// FUSE-NEXT:     %10 = "neura.mul"(%9) {rhs_value = 2 : i32} : (!neura.data<i32, i1>) -> !neura.data<i32, i1>
+// FUSE-NEXT:     %11 = "neura.add"(%10) {rhs_value = 1 : i32} : (!neura.data<i32, i1>) -> !neura.data<i32, i1>
 // FUSE-NEXT:     neura.store_indexed %11 to %8[%nextindex : !neura.data<i64, i1>] !neura.data<memref<?xi32>, i1> : !neura.data<i32, i1>
 // FUSE-NEXT:     neura.ctrl_mov %7 -> %4 : !neura.data<memref<?xi32>, i1> !neura.data<memref<?xi32>, i1>
 // FUSE-NEXT:     neura.ctrl_mov %8 -> %2 : !neura.data<memref<?xi32>, i1> !neura.data<memref<?xi32>, i1>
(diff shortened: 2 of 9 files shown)


tancheng:
> the final operations are reduced to 20, and II = 5 (equals to RecII = 5)

What is the original #ops and II?

include/NeuraDialect/NeuraOps.td
```diff
@@ -172,9 +172,9 @@ def Neura_StoreIndexedOp: Op<NeuraDialect, "store_indexed"> {
 }
 
 // Defines a pointer computation operation.
-def Neura_GEP : Op<NeuraDialect, "gep"> {
+def Neura_GEP : Op<NeuraDialect, "gep", [AttrSizedOperandSegments]> {
```
tancheng: Can u remind me what is `AttrSizedOperandSegments` for?

lib/NeuraDialect/Transforms/Optimizations/HwAgnosticOpt/FoldConstantPass.cpp
```diff
@@ -233,6 +233,89 @@ struct FuseRemRhsConstantPattern : public FuseRhsConstantPattern<neura::RemOp> {
...
+      indices.push_back(operand);
+    }
+
+    // Create new GEP with no base but with const_base_ptr attribute
+    auto fused_gep = rewriter.create<neura::GEP>(
+        gep_op.getLoc(), 
+        gep_op.getResult().getType(),
+        /*base=*/nullptr,
+        indices);
+    addConstantAttribute(fused_gep, "const_base_ptr", base_const_value);
```
tancheng: Can we call it `lhs_arg_value` instead of `const_base_ptr`?

tancheng: Also can you put `TODO: Gather all the attribute -- https://github.com/coredac/dataflow/issues/145` above this line to indicate we need to avoid using string.

lib/NeuraDialect/Transforms/Optimizations/HwAgnosticOpt/FoldConstantPass.cpp
```diff
@@ -233,6 +233,89 @@ struct FuseRemRhsConstantPattern : public FuseRhsConstantPattern<neura::RemOp> {
...
+
+    auto constant_op = dyn_cast<neura::ConstantOp>(addr.getDefiningOp());
+    Attribute addr_const_value = getOriginConstantValue(addr);
+
+    // Create new Store with no addr but with const_dst_ptr attribute
+    auto fused_store = rewriter.create<neura::StoreOp>(
+        store_op.getLoc(),
+        store_op.getValue(),  // Keep the value operand
+        /*addr=*/nullptr);    // Remove addr operand
+    addConstantAttribute(fused_store, "const_dst_ptr", addr_const_value);
```
tancheng: Rename `const_dst_ptr` to `rhs_value`

test/c2llvm2mlir/simple_loop/test.mlir
```diff
@@ -15,3 +15,42 @@
...
+// RUN:   --view-op-graph \
+// RUN:   --architecture-spec=../../arch_spec/architecture.yaml \
+// RUN:   --insert-data-mov \
+// RUN:   --map-to-accelerator="mapping-strategy=heuristic backtrack-config=customized=5,3" %t-kernel.mlir -o %t-kernel-mapped.mlir
+// RUN: FileCheck %s --check-prefix=CHECK-LLVM2NEURA-MAP < %t-kernel-mapped.mlir
+
+// CHECK-LLVM2NEURA: accelerator = "neura"
+// CHECK-LLVM2NEURA: dataflow_mode = "predicate"
+// CHECK-LLVM2NEURA: neura.phi
+// CHECK-LLVM2NEURA: neura.gep
```
tancheng: When checking `neura.gep`, shouldn't we also check whether `rhs_value` exists?

tancheng:
> > > the final operations are reduced to 20, and II = 5 (equals to RecII = 5)
> > 
> > 
> > What is the original #ops and II?
> 
> Original ops are 33, and original II = 7

Cool, then next step is https://github.com/coredac/dataflow/blob/main/test/neura/for_loop/relu_test.mlir, IIRC, its II is 11, let's see whether ur optimization can help it, plz try before and after ur optimization on this test, and include `// CHECK` the updated II in this test. And we will optimize it further with other optimization if necessary.

After above done, let's move to your complex kernel.

test/c2llvm2mlir/simple_loop/test.mlir
```diff
@@ -15,3 +15,55 @@
...
```
tancheng: Can we also check whether there exists `operand_segment_sizes` in this op? which is generated because you use `AttrSizedOperandSegments`.

lib/NeuraDialect/Transforms/Optimizations/HwAgnosticOpt/FoldConstantPass.cpp
```diff
@@ -289,25 +290,25 @@ struct FuseStoreAddrConstantPattern : public OpRewritePattern<neura::StoreOp> {
...
-    // Create new Store with no addr but with const_dst_ptr attribute
+    // Creates new Store with no addr but with rhs_value attribute.
     auto fused_store = rewriter.create<neura::StoreOp>(
         store_op.getLoc(),
-        store_op.getValue(),  // Keep the value operand
-        /*addr=*/nullptr);    // Remove addr operand
-    addConstantAttribute(fused_store, "const_dst_ptr", addr_const_value);
+        store_op.getValue(),  // Keeps the value operand.
+        /*addr=*/nullptr);    // Removes addr operand.
+    addConstantAttribute(fused_store, "rhs_value", addr_const_value);
```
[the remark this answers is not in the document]
YanzhouTang: Let me double confirm, so we only need two addConstantAttribute name : `rhs_value` and `lhs_value` to universally represent all the addConstantAttribute, right? 🤔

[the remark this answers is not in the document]
tancheng: Right. It would help our codegen, as we don't want to specify base/pointer or other fancy attribute in our HW. HW only knows lhs, rhs, or later rename to operand_0/operand_1/operand_2 if we support const folding for fused op with many operands.

tancheng (approved):

Issue #160, opened 2025-10-17
[P0] Problems with e2e conversion


n0thingNoob:
# Summary

Lowering from LLVM/MLIR to the Neura dialect fails on two operations that aren’t currently converted: llvm.fdiv and llvm.fptosi. Additionally, llvm.fdiv receives an operand of type !neura.data<f32, i1>, which triggers a type mismatch.

# Failing IR snippet
%13 = llvm.fdiv %12, %3 : f32
%14 = llvm.fptosi %13 : f32 to i32

# Error messages

- llvm.fdiv op operand #0 must be floating point LLVM type or LLVM dialect-compatible vector of floating point LLVM type, but got !neura.data<f32, i1>.
- No conversion rule exists for llvm.fptosi.

I have uploaded the histogram to the branch testbench. Here is the PR: #161

n0thingNoob:
> Declare the `neura.fdiv` by mimicing:
> 
> [dataflow/include/NeuraDialect/NeuraOps.td](https://github.com/coredac/dataflow/blob/61ea9fe9d03db7078887e8fc786c43d1f6704ef9/include/NeuraDialect/NeuraOps.td#L49-L61)
> 
> Lines 49 to 61 in [61ea9fe](/coredac/dataflow/commit/61ea9fe9d03db7078887e8fc786c43d1f6704ef9)
> 
>  def Neura_RemOp : Op<NeuraDialect, "rem">{ 
>    let summary = "Integer remainder operation"; 
>    let description = [{ 
>      Performs an integer remainder operation, computing the result of 
>      a % b, where % is the remainder operator. 
>   
>      Example: 
>        %result = neura.rem %a, %b : i32 
>    }]; 
>    let arguments = (ins AnyType:$lhs, Optional<AnyType>:$rhs); 
>    let results = (outs AnyType:$result); 
>    let traits = [SameOperandsAndResultElementType]; 
>  } 
> and provide conversion by mimicing:
> 
> [dataflow/lib/Conversion/LlvmToNeura/LlvmToNeuraPass.cpp](https://github.com/coredac/dataflow/blob/61ea9fe9d03db7078887e8fc786c43d1f6704ef9/lib/Conversion/LlvmToNeura/LlvmToNeuraPass.cpp#L122-L135)
> 
> Lines 122 to 135 in [61ea9fe](/coredac/dataflow/commit/61ea9fe9d03db7078887e8fc786c43d1f6704ef9)
> 
[...]

PR #161, opened 2025-10-17
Add histogram testbench files

n0thingNoob:
I am currently working on compiling the testbench file. Here is one I just uploaded: 
- histogram.cpp: Main test program with test data
- histogram_kernel.cpp: Kernel implementation
- histogram_kernel.ll: LLVM IR generated from kernel
- histogram_kernel.mlir: MLIR representation from LLVM IR
- histogram_kernel_neura.mlir: Neura dialect with compilation issues

We have an issue with this transfer. I attached it to the issues https://github.com/coredac/dataflow/issues/160
There is a problem that we might need to solve: 
    %13 = llvm.fdiv %12, %3 : f32
    %14 = llvm.fptosi %13 : f32 to i32
 
There are two ops that are not transferred into the neura operation. I think we might need to add more conversion rules for these Ops. 
For fdiv: 
I got the error message: 'llvm.fdiv' op operand #0 must be floating point LLVM type or LLVM dialect-compatible vector of floating point LLVM type, but got '!neura.data<f32, i1>'
For fptosi: 
We don't have the conversion rules.


test/testbench/histogram/histogram_kernel_neura.mlir
tancheng: You only need to have a `histogram_test.mlir` in this folder and include the command starting with `// RUN: `, and check whatever you wanna check. WDYT?

n0thingNoob: I thought we should have the entire cpp code. So just the kernel's mlir file would be fine?

tancheng: You already have that, right?

Screenshot 2025-10-17 at 4 37 26 PM

n0thingNoob: So only mlir file?

PR #162, opened 2025-10-18
Fix LLVM fdiv and fptosi to Neura conversion and reorganize benchmarks

n0thingNoob:
- Fix Neura_FDivOp and Neura_OrOp type definitions in NeuraOps.td
- Add LlvmFDivToNeuraFDiv and LlvmFPToSIToNeuraCast conversion patterns
- Move CGRA-Bench submodule to test/benchmark/CGRA-Bench
- Add histogram kernel test files to test/e2e/histogram/
I also added the kernel.cpp and kernel.ll to the e2e/histogram because I am not sure if the kernel I am using is correct, which also requires a review.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -83,9 +83,6 @@ jobs:
           -DCMAKE_CXX_FLAGS="-std=c++17"
 
         ninja
-        
-        # Add dataflow tools to PATH
-        echo "${{github.workspace}}/build/bin" >> $GITHUB_PATH
 
         # # install clang-12/opt-12
         # - name: install LLVM and Clang for scripts/experiment


lib/Conversion/LlvmToNeura/LlvmToNeuraPass.cpp
```diff
@@ -134,6 +134,58 @@ struct LlvmSRemToNeuraRem : public OpRewritePattern<LLVM::SRemOp> {
...
+
+struct LlvmFPToSIToNeuraCast : public OpRewritePattern<mlir::LLVM::FPToSIOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(mlir::LLVM::FPToSIOp op,
+                                PatternRewriter &rewriter) const override {
+    Value input = op.getArg();
+    Type result_type = op.getType();
+
+    // Create a cast operation with "fptosi" as the cast type
```
tancheng: `// Creates`

test/e2e/fir/fir_kernel.cpp
tancheng: why this exists? shouldn't we use the CGRA-Bench directly?

test/e2e/histogram/histogram_kernel.ll
tancheng: ditto

test/testbench/FFT/fft.ll
tancheng: why we need this?

test/neura/fusion/kernel.mlir
tancheng: why we need this......

test/c2llvm2mlir/nested_loop/kernel.mlir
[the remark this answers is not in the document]
n0thingNoob: yes, I have removed them in my latest commit

test/testbench/FFT/fft.ll
tancheng: I thought we don't need this `testbench` folder, as you already have `fir` inside the `e2e`. file directory hierarchy:
```
dataflow/test/benchmarks/a.cpp
dataflow/test/benchmarks/b.cpp
dataflow/test/benchmarks/CGRA-Bench

dataflow/test/e2e/a/a_test.mlir
dataflow/test/e2e/fir/fir_test.mlir
dataflow/test/e2e/histogram/histogram_test.mlir
```

test/neura/fusion/kernel.mlir
n0thingNoob: I am working on cleaning it.

test/testbench/FFT/fft.ll
n0thingNoob: Wrongly commit. I used the branch yesterday, and it messed up my directory.

test/e2e/fir/fir_kernel.cpp
tancheng: You basically want to skip the main function, right?
Can you try to put following into the `fir_test.mlir` (try it on your environment first, this is suggested by GenAI)?
```
# 2. Extract *only* your kernel function (e.g., "my_kernel")
#    The --glob flag allows simple name matching.
#    This creates a new file with just "my_kernel" and its dependencies.
llvm-extract --glob="*my_kernel*" -S %t-kernel-full.ll -o %t-kernel-only.ll
```

test/e2e/histogram/histogram.cpp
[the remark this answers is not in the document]
n0thingNoob: Removed

test/e2e/fir/fir_kernel.cpp
n0thingNoob: Does that mean I will need the full code in the same path?

n0thingNoob: got it

n0thingNoob: So, there are a few problems:
The FIR kernel fails with -O2 optimization due to aggressive vectorization because the FIR kernel contains a #pragma clang loop unroll_count(4) vectorize_width(4) directive that forces vectorization even with high optimization levels.
The Histogram Kernel fails with -O0 due to unnecessary type conversions -- llvm.fpext and llvm.fptrunc. 
Do I need to add these two conversions or just leave them there? And how could we solve the vectorization for the FIR?

test/e2e/fir/fir_kernel.ll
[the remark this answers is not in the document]
n0thingNoob: removed

test/e2e/fir/fir_kernel.cpp
tancheng: Would force no vectorize and force no unroll during clang compilation work? Ask GPT.
If it doesn't work, plz let me know the reason, and i can remove those pragma from the test bench.

Histogram, if -O3 works, no need to care -O0? If `llvm.fpext` and `llvm.fptrunc` are blockers anyways, we need to support them.

n0thingNoob: So let me try -O3 for histogram. For the fir, O0 would force no optimization, including vectorization, and it works in the previous version I committed.

tancheng: We have neura.fadd. Why llvm.fadd still there after conversion? Should be a fixable bug in neura world.

tancheng: It is vectorized. Can you try force no vectorization?

test/e2e/fir/fir_kernel.mlir
```diff
@@ -0,0 +1,87 @@
...
+    llvm.store %25, %8 {alignment = 4 : i64} : i32, !llvm.ptr
+    llvm.br ^bb1 {loop_annotation = #loop_annotation}
+  ^bb4:  // pred: ^bb1
+    %26 = llvm.load %9 {alignment = 4 : i64} : !llvm.ptr -> f32
+    %27 = llvm.load %6 {alignment = 8 : i64} : !llvm.ptr -> !llvm.ptr
+    %28 = llvm.getelementptr inbounds %27[%4] : (!llvm.ptr, i64) -> !llvm.ptr, f32
+    llvm.store %26, %28 {alignment = 4 : i64} : f32, !llvm.ptr
+    llvm.return
+  }
+}
```
tancheng: Remove?

test/c2llvm2mlir/nested_loop/kernel.mlir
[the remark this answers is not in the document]
tancheng: These two files are still here?

test/e2e/fir/fir_kernel.mlir
```diff
@@ -0,0 +1,89 @@
+// Compiles the original C kernel to mlir, then lowers it via Neura.
+// Got error when using -O3 -fno-vectorize -fno-slp-vectorize -mllvm -force-vector-width=1
```
tancheng: `// TODO: Got ...`

.github/workflows/main.yml
```diff
@@ -80,6 +83,9 @@ jobs:
           -DCMAKE_CXX_FLAGS="-std=c++17"
 
         ninja
+        
+        # Add dataflow tools to PATH
+        echo "${{github.workspace}}/build/bin" >> $GITHUB_PATH
```
tancheng: What additional "dataflow tools" we need?

test/testbench/Conv2D/.gitkeep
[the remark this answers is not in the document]
tancheng: Cool, thanks~!

.github/workflows/main.yml
n0thingNoob: It should not be dataflow tools. It was added by AI because the workflow failed on GitHub.

tancheng (approved):

PR #166, opened 2025-10-20
Add FMax and FMin Operations Support and More Constant Folding

YanzhouTang:
1. Add support for floating-point maximum (`FMax`) and minimum (`FMin`) operations across the Neura dialect, including operation definitions, LLVM conversions, constant folding, and interpreter support.
2. For PASS `--fold-constant`, add support for FMul, FSub, FMax, FMin, LoadIndexed and StoreIndexed


lib/Conversion/LlvmToNeura/LlvmToNeuraPass.cpp
```diff
@@ -134,6 +134,42 @@ struct LlvmSRemToNeuraRem : public OpRewritePattern<LLVM::SRemOp> {
   }
 };
 
+struct LlvmMaxNumToNeuraFMax : public OpRewritePattern<LLVM::MaxNumOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(LLVM::MaxNumOp op,
```
tancheng: LLVM::MaxNumOp is different LLVM::MaximumOp, can we add attribute to neura.max indicating which one is lowered from? You can also add test for `neura.max()<maxnum>` and `neura.max()<maximum>` when `NAH` is serving as operand.

PR #168, opened 2025-10-21
Update fir int

tancheng:
Provide fir int kernel
 - Update fir kernel test accordingly
 - No alloc needed

@Jackcuii plz check the generated mapping and IR. It has II=5, which is easier for you to parsing. And I update the data type as integer, so you don't need to change types.


YanzhouTang:
Do we need also change the pragma in https://github.com/tancheng/CGRA-Bench/blob/master/kernels/fir/fir.cpp Line37? Or the `#pragma clang loop unroll_count(4) vectorize_width(4)` could make `-fno-vectorize` not take effect.

PR #181, opened 2025-10-29
Update constant folding constraint: only support one constant operand folding for a given operation

YanzhouTang:
As we discuss in #178, our current RTL design only support one constant operand folding for a given operation. 

So if there is an op requires 3 (or more) operands, and 2 (or more) of them are constant, we can only fold 1.

### DIFF
--- a/lib/NeuraDialect/Transforms/Optimizations/HwAgnosticOpt/FoldConstantPass.cpp
+++ b/lib/NeuraDialect/Transforms/Optimizations/HwAgnosticOpt/FoldConstantPass.cpp
@@ -95,9 +95,8 @@ SmallVector<ConstantOperandInfo> analyzeOperandsForFolding(Operation *op) {
         continue;  // Don't fold this one.
       }
       
-      // Rule 2: Special restriction: For operations with 3 or more operands,
-      // only fold at most 1 constant.
-      if (num_operands >= 3 && !constants_to_fold.empty()) {
+      // Rule 2: Only folds at most 1 constant.
+      if (!constants_to_fold.empty()) {
         // Already have one constant to fold, skip the rest.
         continue;
       }
@@ -325,6 +324,7 @@ struct FuseGEPConstantPattern : public GenericFuseConstantPattern<neura::GEP> {
     for (Value val : non_const_operands) {
       // Checks if this is the base operand.
       if (val == orig_base) {
+        assert(num_base == 0);
         operands.push_back(val);
         num_base = 1;
       } else {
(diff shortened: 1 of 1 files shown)


lib/NeuraDialect/Transforms/Optimizations/HwAgnosticOpt/FoldConstantPass.cpp
```diff
@@ -429,6 +434,14 @@ struct FuseStoreIndexedConstantPattern
       return failure();
     }
     
+    // Special restriction: For operations with 3 or more operands,
+    // only fold at most 1 constant (fold value, not base).
+    size_t num_operands = op->getNumOperands();
+    if (num_operands >= 3 && value_is_const && base_is_const) {
+      // Both are constants, but we can only fold one.
+      // Prioritize folding value (lhs), don't fold base (rhs).
```
tancheng: hmm, i thought base is lhs, value is rhs. Can you show the format of store here, so people understand how it looks like?

lib/NeuraDialect/Transforms/Optimizations/HwAgnosticOpt/FoldConstantPass.cpp
```diff
@@ -306,25 +312,24 @@ struct FuseGEPConstantPattern : public GenericFuseConstantPattern<neura::GEP> {
...
-    }
-    
-    for (Value idx : orig_indices) {
-      if (!isOriginConstantOp(idx)) {
-        operands.push_back(idx);
+    for (Value val : non_const_operands) {
+      // Checks if this is the base operand.
+      if (val == orig_base) {
+        operands.push_back(val);
+        num_base = 1;
```
tancheng: Should we add an `ASSERT` before assigning, indicating `assert(num_base == 0)`?

tancheng (approved):

lib/NeuraDialect/Transforms/Optimizations/HwAgnosticOpt/FoldConstantPass.cpp
YanzhouTang: > Should we add an `ASSERT` before assigning, indicating `assert(num_base == 0)`?

Understood. Since the base operand should appear at most once, `num_base` is initialized to 0. When this branch is entered, `num_base` can only be 0 prior to the assignment to 1. Adding the assertion will validate this invariant.

tancheng (approved):

lib/NeuraDialect/Transforms/Optimizations/HwAgnosticOpt/FoldConstantPass.cpp
tancheng: Can you also show an example when `num_operands = 4`? I am just curious the usecase.

YanzhouTang: > Can you also show an example when `num_operands = 4`? I am just curious the usecase.

u can refer to https://github.com/coredac/dataflow/blob/main/test/affine2neura/bert/bert_node28/bert_node28.mlir#70

```
neura.store_indexed %20 to %arg2[%3, %6, %10 : index, index, index] memref<?x128x768xf32> : f32
```

Here `num_operands = 5`

PR #187, opened 2025-11-08
Testbench

n0thingNoob:
Summary
When data moves between PEs via links with register mapping, the generated
ASM instructions use incorrect direction operands. For example, PE(2,1) should
receive data in histogram from PE(2,2) via NORTH direction, but the original code generated
SOUTH instead.

Changes:
1. Fix `handleRegisterRewiring`: Add `invertDir` when computing `incoming_dir`
   for `placeDstDeposit`. The deposit instruction is placed on the destination
   tile and needs to know from which direction data arrives.

2. Fix `handleDirectionRewiring`: Calculate direction based on consumer's actual
   tile location instead of assuming it's always on the link destination tile.
   When consumer is on a different tile, compute the direction from destination
   tile to consumer tile.

Files Changed:
- lib/NeuraDialect/Transforms/GenerateCodePass.cpp
  - Line 519: Add `invertDir` for register deposit direction
  - Lines 535-564: Enhance `handleDirectionRewiring` to handle consumers on
    different tiles
  - Line 593: Pass `topo` parameter to `handleDirectionRewiring`


lib/NeuraDialect/Transforms/GenerateCodePass.cpp
```diff
@@ -515,27 +515,46 @@ struct GenerateCodePass
     if (!links.empty()) {
       // Cross-tile: deposit on destination tile at earliest register ts.
       int dst_tile = topo.dstTileOfLink(links.back().link_id);
-      StringRef incoming_dir = topo.dirFromLink(links.back().link_id);
+      // Computes incoming direction from destination tile's perspective.
+      StringRef incoming_dir = topo.invertDir(topo.dirFromLink(links.back().link_id));
       placeDstDeposit(topo, dst_tile, timestep_0, incoming_dir, register_id, /*asCtrlMov=*/IsCtrl);
 
-      auto cp = operation_placements.lookup(consOp);
+      auto cp = operation_placements.lookup(cons_op);
```
tancheng: replace `cp` with a meaningful name. And instead of `auto`, provide the explicit type.

lib/NeuraDialect/Transforms/GenerateCodePass.cpp
```diff
@@ -570,9 +589,9 @@ struct GenerateCodePass
     }
 
     // Wires each consumer: prefer register rewiring; fallback to direction rewiring.
-    for (auto &[consOp, atVal] : consumers) {
-      if (!handleRegisterRewiring<IsCtrl>(consOp, atVal, regs, links, topo))
-        handleDirectionRewiring<IsCtrl>(consOp, atVal, consumer_direction, links, forwarder);
+    for (auto &[cons_op, at_val] : consumers) {
```
tancheng: Instead of `auto`, provide explicity type. And replace `at_val` with meaningful name.

n0thingNoob: ok

lib/NeuraDialect/Transforms/GenerateCodePass.cpp
```diff
@@ -515,27 +515,46 @@ struct GenerateCodePass
...
+      setConsumerSourceExact(cons_op, at_val, "$" + std::to_string(register_id));
       return true;
     }
     return false;
   }
 
   template<bool IsCtrl>
-  void handleDirectionRewiring(Operation *consOp, Value atVal, StringRef consumer_direction,
-                               const SmallVector<LinkStep, 8> &links, Operation *forwarder) {
+  void handleDirectionRewiring(Operation *cons_op, Value at_val, StringRef consumer_direction,
```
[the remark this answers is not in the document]
n0thingNoob: Ok

test/e2e/bicg/bicg_kernel.mlir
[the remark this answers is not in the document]
n0thingNoob: Ok, let's try that after the ISCA

Issue #188, opened 2025-11-08
[P0] Support constant folding for shl


tancheng:
Based on the comment: https://github.com/coredac/dataflow/pull/187#discussion_r2506328987

PR #192, opened 2025-11-22
Add constant folding support for Shl, And, Or bitwise operations

YanzhouTang:
As we discussed in #188 , we add more operations supported by `--fold-constant` PASS.


tancheng (approved):

tancheng:
Excellent~ thanks @YanzhouTang!

PR #228, opened 2026-01-03
Update the code generate pass

n0thingNoob:
fix the problem in this https://github.com/sarchlab/Zeonica-issues/issues/16

### DIFF
--- a/test/code_gen/test_code_generate.mlir
+++ b/test/code_gen/test_code_generate.mlir
@@ -144,11 +144,11 @@ func.func @loop_test() -> f32 {
 // ASM:      PE(1,1):
 // ASM-NEXT: {
 // ASM-NEXT:   GRANT_PREDICATE, [WEST, RED], [EAST, RED] -> [SOUTH, RED] (t=7, inv_iters=1)
+// ASM-NEXT:   DATA_MOV, [EAST, RED] -> [WEST, RED] (t=7, inv_iters=1)
 // ASM-NEXT:   DATA_MOV, [NORTH, RED] -> [$1] (t=7, inv_iters=1)
 // ASM-NEXT: } (idx_per_ii=2)
 // ASM-NEXT: {
 // ASM-NEXT:   GRANT_PREDICATE, [$0], [$1] -> [$0] (t=8, inv_iters=1)
-// ASM-NEXT:   DATA_MOV, [EAST, RED] -> [WEST, RED] (t=8, inv_iters=1)
 // ASM-NEXT: } (idx_per_ii=3)
 // ASM-NEXT: {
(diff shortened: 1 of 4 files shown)


n0thingNoob:
So I still have a problem with this ir:
%25 = "neura.icmp"(%24) <{cmpType = "eq"}> {dfg_id = 20 : i32, mapping_locs = [{id = 11 : i32, index_per_ii = 3 : i32, invalid_iterations = 0 : i32, resource = "tile", time_step = 3 : i32, x = 3 : i32, y = 2 : i32}], rhs_value = 32 : i64} : (!neura.data<i64, i1>) -> !neura.data<i1, i1>
 
%35 = "neura.data_mov"(%25) {dfg_id = 23 : i32, mapping_locs = [{id = 352 : i32, index_per_ii = 3 : i32, invalid_iterations = 0 : i32, per_tile_register_id = 0 : i32, resource = "register", time_step = 3 : i32}, {id = 35 : i32, index_per_ii = 4 : i32, invalid_iterations = 0 : i32, resource = "link", time_step = 4 : i32}, {id = 31 : i32, index_per_ii = 0 : i32, invalid_iterations = 1 : i32, resource = "link", time_step = 5 : i32}, {id = 289 : i32, index_per_ii = 1 : i32, invalid_iterations = 1 : i32, per_tile_register_id = 1 : i32, resource = "register", time_step = 6 : i32}]} : (!neura.data<i1, i1>) -> !neura.data<i1, i1>


%36 = neura.grant_predicate %34, %35 {dfg_id = 37 : i32, mapping_locs = [{id = 9 : i32, index_per_ii = 2 : i32, invalid_iterations = 1 : i32, resource = "tile", time_step = 7 : i32, x = 1 : i32, y = 2 : i32}]} : !neura.data<i32, i1>, !neura.data<i1, i1> -> !neura.data<i32, i1>
   
What is the first register in %35 do? I didn't see any other tile use this value in the reg

n0thingNoob:
> Stay in Tile A reg, then go to another tile, then stay in another tile B reg?

But since it would not be used again, we do not need to show this Tile A reg in our YAML file, right? I just want to make sure it would not affect our final output.

ShangkunLi:
> > Stay in Tile A reg, then go to another tile, then stay in another tile B reg?
> 
> But since it would not be used again, we do not need to show this Tile A reg in our YAML file, right? I just want to make sure it would not affect our final output.

The mapped ir means we need to store the result produced by `icmp` in Tile A for one cycle, then send it to Tile B by links.

And the compiler is responsible for register allocation. If we remove this from YAML file, how could we know where to store (in which tile and which local register) the result produced by `icmp`?

n0thingNoob:
I need to revise our logic

PR #229, opened 2026-01-04
Revise the Generate Code Pass

n0thingNoob:
tmp-generated-dfg
Fixed the problem before in the issue https://github.com/sarchlab/Zeonica-issues/issues/16


n0thingNoob:
> Can you explain why there was an issue before, and how this PR resolves the issue? I originally thought this is a mapping issue, but it seems it is purely a codegen issue?

Yes, so I revise the data_mov part but forgot to change the ctrl_mov as well.

PR #233, opened 2026-01-09
Add General Test for Taskflow Dialect

ShangkunLi:
Hi~ @tancheng,

I am trying to implement the ConvertAffineToTaskflow pass this week. The problem that I encountered these days is that we cannot exhaust all the affine structures in this pass. I have already written a 1.7K-line conversion pass to convert the following two cases (`multi-nested` and `irregular-loop`) successfully.

But when I try to add a new case, the conversion pass cannot process that structure, and more specific handling code needs to be added.

For linalg dialect, it’s a good idea to implement such a pass as there are only data dependencies between different tasks. However, for affine.for (especially for imperfect-nested loops), the nested structures are too hard for us to analyze.

So in this pr, I just put the transformed ir of `multi-nested` and `irregular-loop` in the tests. Just to make sure that the defined dialect is correct and can represent such structures.

More discussions are needed for converting from high-level representations.


tancheng (commented):
Can you show a unsupported example? I thought we anyway can represent any example with a naive way (w/o analyzing dependency, i.e., assume all data are dependent).

ShangkunLi:
Re-defined the `taskflow` computation abstraction, so close this pr. Will create a new pr soon~

PR #235, opened 2026-01-13
Introducing task level computation abstraction

ShangkunLi:
In this pr, 
1. re-define the computation of `taskflow.task`: we ensure there are only data dependencies (e.g., producer-consumer, RAW, WAR, WAW) between different tasks
2. introduce `taskflow.counter` op: convert all the loop control into counter tree/graph, each counter represents a loop control
3. introduce `taskflow.hyperblock`: wrap the rest code into `hyperblock`

This design enables us to perform three-level optimizations:
1. Pre-task optimization: before transforming to taskflow, we can perform high-level optimizations to explore the parallelism
2. Resource-agnostic optimization: perform task fission/fusion to minimize the data dependencies between different tasks
3. Resource-aware optimization: task legalization & resource binding (e.g., one task on multiple CGRAs, fuse multiple task to maximize resource utilization)

TODO:
Hyperblock fusion to generate the DFG/integrate to the `neura.kernel` op.


ShangkunLi:
> * Bufferization is performed between task and hyperblock IR? I saw affine becomes memref.

When we extract `affine.load/store/if` operation into `hyperblock`, we cannot preserve the map relationtiops in the hyperblock region. So I use
```
// Creates a taskflow.hyperblock operation from HyperblockInfo.
static TaskflowHyperblockOp createHyperblock(
    OpBuilder &builder, Location loc, const HyperblockInfo &info,
    Block *task_body,
    const DenseMap<affine::AffineForOp, LoopInfo *> &loop_info_map) {
  ...
  ...
  MLIRContext *context = hyperblock_op.getContext();
  RewritePatternSet patterns(context);

  populateAffineToStdConversionPatterns(patterns);
  ConversionTarget target(*context);
  target.addLegalDialect<arith::ArithDialect, memref::MemRefDialect,
                         func::FuncDialect, taskflow::TaskflowDialect>();
  target.addIllegalOp<affine::AffineLoadOp, affine::AffineStoreOp,
                      affine::AffineIfOp>();
  if (failed(applyPartialConversion(hyperblock_op, target, std::move(patterns)))) {
    assert(false && "Affine to Standard conversion failed.");
  }

  return hyperblock_op;
}
```
to convert `affine.load/store/if` operations into standard operations (e.g., `scf.if`, `memref.load/store`)

> * `hyperblock` is region but it doesn't have yield?

I use the trait `SingleBlockImplicitTerminator<"TaskflowHyperblockYieldOp">` in `taskflow.yield`. This means if the hyperblock has an output, the `taskflow.hyperblock.yield` exists explicitly, otherwise implicitly. 

> * Why is it called `hyperblock`? It sounds like just `block`.

Because actually, we can also wrap the innermost loop (including the loop control) into the hyperblock, and we can still handle loop control w/o counters. Furthermore, if a loop body contains `if-else` conditions, we need to lower them to CDFG-based ir to build the DFG.

PR #241, opened 2026-01-18
[Refactor] Introduce NeuraAttributes.h to manage attribute constants

shiyunyao:
This PR addresses part of #145 by introducing a centralized header NeuraAttributes.h and replacing a subset of hardcoded magic strings in the lib/ directory.


tancheng:
Can we also have following in the header file? https://github.com/coredac/dataflow/blob/2a6fdfa3375882ead687cda895002804c96094b3/lib/NeuraDialect/Transforms/GenerateCodePass.cpp#L204-L224

And anything introduced from https://github.com/coredac/dataflow/pull/157/changes

PR #242, opened 2026-01-18
Enable task canonicalization pass in task flow dialect

ShangkunLi:
In this pr:
- fix some bugs in `convert-affine-to-taskflow` and `construct-hyperblock-from-task` passes
- create a new pass called `canonicalize-task` in taskflow dialect

This new canonicalization pass outputs the canonicalized task. There is only one hyperblock in each task and its corresponding counters.

Based on the canonicalized task graph, we can further perform task fusion/fission and resource binding.

The next step is to:
1. Determine the design of our affine controller
2. Enable mapping a task on a CGRA

### DIFF
--- a/lib/TaskflowDialect/Transforms/ConstructHyperblockFromTaskPass.cpp
+++ b/lib/TaskflowDialect/Transforms/ConstructHyperblockFromTaskPass.cpp
@@ -203,7 +203,7 @@ static void extractHyperblocksInfoFromRegion(
       SmallVector<Value> loop_indices = parent_indices;
       loop_indices.push_back(loop_info->counter_index);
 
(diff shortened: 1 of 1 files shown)


tancheng (approved):

lib/TaskflowDialect/Transforms/ConstructHyperblockFromTaskPass.cpp
```diff
@@ -205,12 +203,51 @@ static void extractHyperblocksInfoFromRegion(
       SmallVector<Value> loop_indices = parent_indices;
       loop_indices.push_back(loop_info->counter_index);
 
+      // 分析哪些 current_ops 被这个循环使用
```
guosran: plz use English here

PR #252, opened 2026-01-29
Add function (task)-level hardware target assignment pass for heterogeneous computing

YanzhouTang:
## Summary

This PR introduces a new pass `AssignTaskTarget` that operates at a higher abstraction level than existing partitioning mechanisms. It assigns hardware targets (CPU, CGRA, DOE) to compute functions before they are lowered to taskflow operations, enabling coarse-grained workload partitioning in heterogeneous computing systems.

## Motivation

The existing `PartitionTaskByTarget` pass operates at the taskflow level (loop-to-CGRA mapping), which is fine-grained for certain use cases. We need a higher-level pass that can:

1. **Partition at function granularity**: Assign entire computational functions to different hardware units
2. **Early hardware mapping**: Make hardware decisions at the Linalg/func level before lowering to lower-level IRs
3. **Enable heterogeneous orchestration**: Support scenarios where different components run on different hardware accelerators (Now we support NeRF algorithm)

### New Pass: `AssignTaskTarget`

- **Location**: `lib/Conversion/AssignTaskTarget/`
- **Functionality**: 
  - Analyzes function names and assigns `target.device` attributes
  - Supports CPU, CGRA, and DOE (Data Orchestration Engine) hardware targets
  - Uses pattern matching for initial implementation (extensible to more sophisticated analysis)
- **Usage**: `mlir-neura-opt --assign-task-target input.mlir`

**Example transformation:**
// Before
func.func @hash_encoder_func(...) { ... }

// After
func.func @hash_encoder_func(...) attributes {target.device = "doe"} { ... }

### DIFF
--- a/test/multi-cgra/taskflow/nerf_hash_grid/build_modular_mlir.py
+++ b/test/multi-cgra/taskflow/nerf_hash_grid/build_modular_mlir.py
@@ -128,9 +128,9 @@ def extract_and_rename_function(mlir_str, new_name):
   for line in lines:
     if 'func.func @forward(' in line:
       in_function = True
-      # Rename and mark as private.
+      # Rename function
       line = line.replace('func.func @forward',
-                          f'func.func private @{new_name}')
+                          f'func.func @{new_name}')
 
     if in_function:
       func_lines.append(line)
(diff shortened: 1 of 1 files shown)


lib/Conversion/AssignTaskTarget/AssignTaskTargetPass.cpp
```diff
@@ -0,0 +1,133 @@
...
+//===----------------------------------------------------------------------===//
+
+/// Determines the hardware target for a given function based on its name.
+/// This function implements a simple pattern-matching strategy:
+///   - ray_sampler* -> CPU
+///   - hash_encoder* -> DOE
+///   - nerf_mlp* -> CGRA
+///   - nerf_forward (top-level) -> CPU
+///   - default -> CPU
+static StringRef matchHardwareTarget(StringRef funcName) {
```
HobbitQia: I would like to know if this function could be extended to accept additional files as input. This would facilitate the assignment of tasks to various hardware platforms without the need for hard-coded configurations.

tancheng:
what is the principle to determine which accelerator should be assigned to which kernel?

PR #247, opened 2026-01-24
Enable lowering from taskflow.task to neura.kernel & Mapping

ShangkunLi:
Sorry for this too large pr......
# Counter Classification
We classify the counter into three types:
1. `root`: no parent, has child(ren)
2. `relay`: has parent, has child(ren)
3. `leaf`: has parent, no child

We need to map each counter op onto the tile array. But only `leaf` counter has the self-increment logic in FU. For other two types, they only have a register to store the counter values, the values are updated through off-array affine controller.

# Task Classification
We classify tasks into two categories:
1. task with `taskflow.counter`:
```
module {
  func.func @_Z6kernelPiS_S_(%arg0: memref<?xi32>, %arg1: memref<?xi32>, %arg2: memref<?xi32>) -> i32 attributes {llvm.linkage = #llvm.linkage<external>} {
    %c0_i32 = arith.constant 0 : i32
    %value_outputs = "taskflow.task"(%arg0, %arg2, %c0_i32) <{operandSegmentSizes = array<i32: 2, 1>, resultSegmentSizes = array<i32: 0, 1>, task_name = "Task_0"}> ({
    ^bb0(%arg3: memref<?xi32>, %arg4: memref<?xi32>, %arg5: i32):
      %0 = taskflow.counter attributes {counter_id = 0 : i32, counter_type = "leaf", lower_bound = 0 : index, step = 1 : index, upper_bound = 32 : index} : index
      %1 = "taskflow.hyperblock"(%0, %arg5) <{operandSegmentSizes = array<i32: 1, 1>}> ({
      ^bb0(%arg6: index, %arg7: i32):
        %2 = memref.load %arg3[%arg6] : memref<?xi32>
        %3 = memref.load %arg4[%arg6] : memref<?xi32>
        %4 = arith.muli %2, %3 : i32
        %5 = arith.addi %arg7, %4 : i32
        taskflow.hyperblock.yield iter_args_next(%5 : i32) results(%5 : i32)
      }) : (index, i32) -> i32
      "taskflow.yield"(%1) <{operandSegmentSizes = array<i32: 0, 1>}> : (i32) -> ()
    }) : (memref<?xi32>, memref<?xi32>, i32) -> i32
    return %value_outputs : i32
  }
}
```
This kind of task is **driven by the counter**; it is also terminated by the (`root`) counter (`leaf` counter when there is only one counter).

This kind of task can be further classified into two categories:
[...]

PR #259, opened 2026-02-01
Enable Atomic Canonical Task Generation

ShangkunLi:
This pr enables the atomic canonical task generation, with the following specific modifications:
- Enable the `affine-loop-tree-serialization` pass to build the atomic affine loop tree (minimized canonicalized task) as described in #247 
- Enhance the `taskflow.task` definition with its original read/write memrefs (This is used for multi-cgra Memory-Task Co-Mapping @guosran)
- Enhance the `create-hyperblock-from-task` with the ability to detect prologue-loop-epilogue pattern:
```
// Prologue-Loop-Epilogue Code means code that appears before and after an inner
// loop. Example: for %i (outer loop) {
//   <prologue code> ← Prologue Code
//   for %j (nested loop) {
//     <loop body>
//   }
//   <epilogue code>  ← Epilogue Code
// }
```
If either prologue code or epilogue code exists, we will wrap all the inner code as a hyperblock:
```
//   taskflow.hyperblock{
//   <prologue code> ← Prologue Code
//   for %j (nested loop) {
//     <loop body>
//   }
//   <epilogue code>  ← Epilogue Code
//   }
```

@guosran, after getting the atomic canonical task, this could be a baseline for our multi-cgra mapping. Please sync with this.
The next pr is to enable wrapping only the innermost loop as a kernel, which is a start point for @HobbitQia.

### DIFF
--- a/lib/TaskflowDialect/Transforms/Optimizations/AffineLoopTreeSerializationPass.cpp
+++ b/lib/TaskflowDialect/Transforms/Optimizations/AffineLoopTreeSerializationPass.cpp
@@ -194,7 +194,7 @@ public:
         loop_builder = OpBuilder::atBlockEnd(current_insert_block);
       }
 
-      // Prepare iter_args for the new loop.
+      // Prepares iter_args for the new loop.
       SmallVector<Value> iter_args_init_values;
       if (node->loop_op.getNumIterOperands() > 0) {
         for (Value init : node->loop_op.getInits()) {
@@ -202,17 +202,17 @@ public:
         }
       }
 
-      // Create new loop with same bounds and iter_args.
+      // Creates new loop with same bounds and iter_args.
       auto new_loop = loop_builder.create<affine::AffineForOp>(
           loc, node->lower_bound, node->upper_bound, node->step,
           iter_args_init_values);
 
       created_loops.push_back(new_loop);
 
-      // Map the old induction variable to the new one.
+      // Maps the old induction variable to the new one.
       mapping.map(node->loop_op.getInductionVar(), new_loop.getInductionVar());
 
-      // Map the old iter_args (block args) to the new iter_args (block args).
+      // Maps the old iter_args (block args) to the new iter_args (block args).
       if (node->loop_op.getNumRegionIterArgs() > 0) {
         for (auto [old_arg, new_arg] :
              llvm::zip(node->loop_op.getRegionIterArgs(),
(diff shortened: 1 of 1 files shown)


ShangkunLi:
> > > So there is no inner hyper-block any more?
> > 
> > 
> > Only the perfect nested loop has one inner hyperblock (w/o any loops). For imperfect nested loops, we need to wrap the prologue, epilogue, and the nested loops into a hyperblock.
> > We can use the `--affine-loop-perfection` pass to transform loops into perfect nested loops as much as possible. Combined with the `--affine-loop-tree-serialization`, we can transform loops into multiple perfect nested affine loops without sibling loops nested inside.
> 
> So the perfect nested-loop would have nested hyperblock? Can you somehow show different cases using figures in this PR's description?

Screenshot 2026-02-02 at 15 24 19

As shown in the figure,
1. If we encounter an imperfect nested loop w/o any preprocessing, we will only transform the **perfect nested band** into counters and wrap the rest of the imperfect nested part into a hyperblock. This enables we can transform any input task into a canonical task `<counter + one hyperblock>`.
2. After applying an `--affine-loop-tree-serialization`, we can eliminate sibling loops. To ensure each task has minimal control logic. 
3. After applying an `--affine-loop-perfection` (will make a pr soon), we can minimize the loop-based control logic in each hyperblock.
4. Applying both `--affine-loop-tree-serialization` and `--affine-loop-perfection` generates the atomic canonical task `<minimal control logic in each task + minimal control logic in each hyperblock>`

p.s. A **perfect nested band** is a sequence of perfectly nested loops (each loop has exactly one child loop w/o any prologue/epilogue code)

ShangkunLi:
A more robust `--construct-hyperblock-from-task` will be needed. A good idea is to detect the maximum perfect nested `affine.for`s and transform them into `taskflow.counters`, then wrap the inner non-perfect part as a hyperblock driven by these counters.

I will update the pass implementation ASAP.

lib/TaskflowDialect/Transforms/Optimizations/AffineLoopTreeSerializationPass.cpp
```diff
@@ -0,0 +1,439 @@
...
+        for (Value operand : original_yield->getOperands()) {
+          yielded_values.push_back(mapping.lookupOrDefault(operand));
+        }
+        leaf_yield_builder.create<affine::AffineYieldOp>(loc, yielded_values);
+      } else {
+        // Should not happen for valid AffineForOp
+        OpBuilder leaf_yield_builder =
+            OpBuilder::atBlockEnd(new_leaf.getBody());
+        leaf_yield_builder.create<affine::AffineYieldOp>(loc);
+      }
```
guosran: What about adding an assertion here?

ShangkunLi:
> So hyperblock is recognized based on imperfect loop.

Yes, we wrap the deepest perfect nested band's loop body as a hyperblock.

> BTW, I saw you embed `accelerator` to `func.func`. Plz sync with @YanzhouTang that whether we need to assign accelerator to `task` or assign accelerator to `func`.

I have discussed with @YanzhouTang yesterday, we agreed that the tha accelerator assignment should be performed on the `func.func` level.

lib/TaskflowDialect/Transforms/Optimizations/AffineLoopTreeSerializationPass.cpp
ShangkunLi: Good idea~ Updated~

PR #261, opened 2026-02-03
Wrap Innermost Loop as neura.kernel

ShangkunLi:
Hi~ @HobbitQia, I enhance the `convert-taskflow-to-neura` pass with two modes:
- 'hyperblock' mode: this mode runs the flow with counters
- 'innermost' mode: this mode wraps each innermost loop as the `neura.kernel`; this could be a starting point for your fusion

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,7 @@ build/
 /test/**/*.yaml
 !arch_spec_example.yaml
 !architecture.yaml
-!architecture_no_counter.yaml
+!architecture_with_counter.yaml
 /test/**/*.asm
 /test/**/*.json
 /test/**/*.dot
--- a/test/arch_spec/architecture.yaml
+++ b/test/arch_spec/architecture.yaml
@@ -15,7 +15,7 @@ per_cgra_defaults:
 
 tile_defaults:
   num_registers: 32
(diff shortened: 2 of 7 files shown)


tancheng:
I still think `mode` is unnecessary. We should have `support_counter` in `arch_spec.yaml` to determine which mode we are targeting. Instead of using different flag. How does this sound?

tancheng:
So you mean `mode=innermost` will have no `hyperblock` exist?

The principle should be:
 - `neura-compiler` and `mlir-neura-opt` should have the same functionality. `opt` is just pick some passes for verification. we should have single compiler no matter what use case. All passes should be applied on different inputs, just some passes might not make impact on the IRs.

So both opt and compiler should have all 3 passes enabled:
```
--convert-affine-to-taskflow
--contruct-hyperblock-from-task
--convert-taskflow-to-neura
```

And `hyperblock` should exist in both cases. Is this possible?

test/arch_spec/architecture_no_counter.yaml
tancheng: Can we name this as `architecture.yaml`, while the one with counter can be named as `architecture_with_counter.yaml`?

BTW, I didn't see the one with counter is uploaded.

PR #266, opened 2026-02-09
Handle memref.copy/subview ops & Add compiler pass pipeline

ShangkunLi:
In this pr:
1. Add two passes to handle the `memref.copy` and `memref.subview` ops
For `memref.subview` ops, we transform such ops into `affine.load/store` directly, like:
```mlir
// With memref.subview op
func.func @example(%arg0: memref<10x20xf32>) {
  %subview = memref.subview %arg0[2, 3][4, 5][1, 1] 
      : memref<10x20xf32> to memref<4x5xf32, strided<[20, 1], offset: 43>>
  
  affine.for %i = 0 to 4 {
    affine.for %j = 0 to 5 {
      %val = affine.load %subview[%i, %j] : memref<4x5xf32, strided<[20, 1], offset: 43>>
      // use %val
    }
  }
  return
}

// memref.subview folded
func.func @example(%arg0: memref<10x20xf32>) {
  affine.for %i = 0 to 4 {
    affine.for %j = 0 to 5 {
      // Index adjustment: [%i, %j] + offset [2, 3] = [%i+2, %j+3]
      %val = affine.load %arg0[%i + 2, %j + 3] : memref<10x20xf32>
      // use %val
    }
  }
  return
}
```

For `memref.copy` ops, we transform it to `affine.for` with `affine.load/store`, like:
```mlir
// With memref.copy
func.func @copy_example(%src: memref<8x8xf32>, %dst: memref<8x8xf32>) {
  memref.copy %src, %dst : memref<8x8xf32> to memref<8x8xf32>
  return
}

// After transformation
func.func @copy_example(%src: memref<8x8xf32>, %dst: memref<8x8xf32>) {
  affine.for %i = 0 to 8 {
    affine.for %j = 0 to 8 {
      %val = affine.load %src[%i, %j] : memref<8x8xf32>
      affine.store %val, %dst[%i, %j] : memref<8x8xf32>
    }
  }
  return
}
```
2. Add two compiler pass pipeline:
- `--tosa-to-affine-conversion`: lowers from `tosa` dialect to `affine` dialect
- `--taskflow-conversion`: converts affine dialects to `taskflow.task` + `neura.kernel`

### DIFF
--- a/lib/TaskflowDialect/TaskflowPasses.cpp
+++ b/lib/TaskflowDialect/TaskflowPasses.cpp
@@ -17,11 +17,12 @@
 using namespace mlir;
 using namespace mlir::bufferization;
 
-// This pass pipeline can convert all the other dialects into the Neura dialect.
+// This pass pipeline can convert affine dialect into taskflow dialect with
+// neura.kernel op.
 void mlir::taskflow::registerTaskflowConversionPassPipeline() {
   PassPipelineRegistration<>(
       "taskflow-conversion",
-      "Convertes affine dialects to taskflow dialect with neura.kernel ops.",
+      "Converts affine dialects to taskflow dialect with neura.kernel ops.",
       [](OpPassManager &pm) {
         pm.addPass(mlir::createConvertAffineToTaskflowPass());
         pm.addPass(mlir::taskflow::createConstructHyperblockFromTaskPass());


ShangkunLi:
> affine.load/store will later be lowered to what?

They will be handled by the neura lowering process. They can be either lowered to `neura.load_index/stroe_idnex`, or lowered to `llvm.gep` + `llvm.load/store`.

tancheng (approved):

Issue #283, opened 2026-03-15
[P1] Register cluster constraint during mapping


tancheng:
If the `reg -> outport`'s reg is in the same `reg cluster` as the computation's input reg, it should be the identical register.

From https://github.com/coredac/dataflow/issues/278.

guosran:
Hi I’m traveling, and I will be available at nights ^^

________________________________

PR #286, opened 2026-03-17
Data Move Register Bypass

guosran:
## Summary
This PR addresses issue #283 by enforcing the register-bank (RegisterFile) read constraint directly in the mapper’s global availability check. When a routing bypass (MOV) and a computation both read from the same register cluster on the same tile at the same time slot, they are forced to use the identical intra-index register, while MOV-only traffic remains unconstrained.

## Changes
- Updated `MappingState::isAvailableAcrossTime` in `MappingState.cpp` to detect `MOV` vs `compute` accesses within the same RegisterFile and reject conflicting mixed accesses across different sibling registers in the same time slot.
- Kept behavior unchanged for different register clusters and for MOV-only cases.

### DIFF
--- a/lib/NeuraDialect/Transforms/GenerateCodePass.cpp
+++ b/lib/NeuraDialect/Transforms/GenerateCodePass.cpp
@@ -345,7 +345,7 @@ struct GenerateCodePass
 
   StringRef getArgument() const override { return "generate-code"; }
   StringRef getDescription() const override {
-    return "CGRA YAML/ASM gen (multi-hop routers + endpoint register deposit + timing-aware rewiring, with CTRL_MOV kept). DATA_MOV reg->outport uses hardware bypass (no egress instruction).";
+    return "CGRA YAML/ASM gen (multi-hop routers + endpoint register deposit + timing-aware rewiring, with CTRL_MOV kept). Validates single-read-port constraint for reg->outport bypass.";
   }
   void getDependentDialects(DialectRegistry &registry) const override {
     registry.insert<mlir::func::FuncDialect>();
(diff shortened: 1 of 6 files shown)


tancheng:
Shouldn't we work on mapping algorithm? i.e., sth like: https://github.com/coredac/dataflow/blob/c5e1f508ff06ae88a62aa5f495e240a53b7aafc5/lib/NeuraDialect/Mapping/mapping_util.cpp#L594

lib/NeuraDialect/Mapping/mapping_util.cpp
```diff
@@ -551,14 +551,43 @@ bool mlir::neura::tryRouteBackwardMove(Operation *mov_op, MappingLoc src_loc,
 Register *mlir::neura::getAvailableRegister(const MappingState &state,
                                             Tile *tile, int start_time,
                                             int exclusive_end_time) {
+  // Collect registers occupied by computations (non-MOV operations) on this
```
tancheng: `// Collects`

lib/NeuraDialect/Mapping/mapping_util.cpp
```diff
@@ -551,14 +551,43 @@ bool mlir::neura::tryRouteBackwardMove(Operation *mov_op, MappingLoc src_loc,
...
-    // input direction could be active at one time).
-    if (state.isAvailableAcrossTimeInRange(reg, start_time,
-                                           exclusive_end_time)) {
-      return reg;
+    for (int t = start_time; t < exclusive_end_time; ++t) {
+      auto op = state.getOpAt({reg, t});
+      if (!op.has_value()) continue;
+      Operation *mapped_op = op.value();
+      if (isa<neura::DataMovOp>(mapped_op) || isa<neura::CtrlMovOp>(mapped_op))
+        continue;
```
tancheng: ditto

lib/NeuraDialect/Mapping/mapping_util.cpp
```diff
@@ -551,14 +551,43 @@ bool mlir::neura::tryRouteBackwardMove(Operation *mov_op, MappingLoc src_loc,
...
+      computation_regs.insert(reg);
+    }
+  }
+
+  for (Register *reg : tile->getRegisters()) {
+    if (!state.isAvailableAcrossTimeInRange(reg, start_time,
```
tancheng: Instead of collects the registers for non-MOV ops first, can we just update the `isAvailableAcrossTime()`? https://github.com/coredac/dataflow/blob/c5e1f508ff06ae88a62aa5f495e240a53b7aafc5/lib/NeuraDialect/Mapping/MappingState.cpp#L118

checking whether it is a register, and whether the register's cluster has any other register is occupied for either mov or compute?

PR #293, opened 2026-03-19
Register Bypass Constraint

guosran:
## Summary
This PR addresses issue #283 by enforcing the register-bank (RegisterFile) read constraint directly in the mapper’s global availability check. When a routing bypass (MOV) and a computation both read from the same register cluster on the same tile at the same time slot, they are forced to use the identical intra-index register, while MOV-only traffic remains unconstrained.

## Changes
- Updated `MappingState::isAvailableAcrossTime` in `MappingState.cpp` to detect `MOV` vs `compute` accesses within the same RegisterFile and reject conflicting mixed accesses across different sibling registers in the same time slot.
- Kept behavior unchanged for different register clusters and for MOV-only cases.

### DIFF
--- a/include/NeuraDialect/Mapping/MappingState.h
+++ b/include/NeuraDialect/Mapping/MappingState.h
@@ -14,11 +14,10 @@ namespace neura {
 
 // Occupy status for multi-cycle pipeline support.
 // These states define how a tile/FU is occupied at a given time step.
-#define SINGLE_OCCUPY 0     // A single-cycle op is in the FU (exclusive)
+#define SINGLE_OCCUPY     0 // A single-cycle op is in the FU (exclusive)
 #define START_PIPE_OCCUPY 1 // A multi-cycle op starts in the FU
-#define END_PIPE_OCCUPY 2   // A multi-cycle op ends in the FU
-#define IN_PIPE_OCCUPY 3    // A multi-cycle op is occupying the FU (pipelined)
-
+#define END_PIPE_OCCUPY   2 // A multi-cycle op ends in the FU
+#define IN_PIPE_OCCUPY    3 // A multi-cycle op is occupying the FU (pipelined)
 // Represents a spatial-temporal location: (resource, time_step)
 struct MappingLoc {
   BasicResource *resource;
@@ -86,7 +85,7 @@ public:
   // For example, if II is 4, and we want to check (tile 2, step 5), then
   // it will check (tile 2, step 1), (tile 2, step 5), (tile 2, step 9), etc.
   bool isAvailableAcrossTime(const MappingLoc &loc,
-                             neura::DataMovOp op = {}) const;
+                             neura::DataMovOp move_op = nullptr) const;
 
   // Checks if a location is available for a specific occupy status.
   // This implements the pipeline-aware availability checking:
@@ -95,7 +94,7 @@ public:
   // - END_PIPE_OCCUPY: available if free or IN_PIPE_OCCUPY or START_PIPE_OCCUPY
   // - IN_PIPE_OCCUPY: always available (can pipeline with any status)
   bool isAvailableForOccupyStatus(const MappingLoc &loc, int new_occupy_status,
-                                  neura::DataMovOp op = {}) const;
+                                  neura::DataMovOp move_op = nullptr) const;
 
   // Gets the occupy status at a specific location across time domain.
   // Returns -1 if the location is not occupied.
@@ -106,7 +105,7 @@ public:
   // time step.
   bool isAvailableAcrossTimeInRange(BasicResource *resource, int start_time,
                                     int exclusive_end_time,
-                                    neura::DataMovOp op = {}) const;
+                                    neura::DataMovOp move_op = nullptr) const;
 
   // Checks availability of a register's cluster write port across the relevant
   // time steps.  Returns false if a DIFFERENT register in the same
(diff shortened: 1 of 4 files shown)


lib/NeuraDialect/Mapping/MappingState.cpp
```diff
@@ -116,35 +116,186 @@ void MappingState::unbindOp(Operation *op) {
...
+        auto violatesClusterReadConstraintAt = [&](int t) -> bool {
+          // Tracks whether sibling registers in the file are used by mov or compute in this slot.
+          bool otherRegHasMov = false;
+          bool otherRegHasCompute = false;
+
+          for (const auto &[_, sibling] : reg_file->getRegisters()) {
+            if (sibling == reg) {
+              continue;
+            }
+            auto itOcc = occupied_locs.find({sibling, t});
```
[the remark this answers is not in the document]
ShangkunLi: Yes, I think we can maintain a record table to get the occupation info faster.

For example, introducing the `reg_file_occupy_records` in the `MappingState` class.

```C++
struct RegClusterOccupyStatus {
  int mov_count = 0;
  int compute_count = 0;

  bool hasConflict() const {
    return mov_count > 0 && compute_count > 0;
  }
};

class MappingState {
// Existing Code ....
std::unordered_map<RegisterFile*, std::unordered_map<int, RegClusterOccupyStatus>> reg_file_occupy_records;`
}
``

File: include/NeuraDialect/Mapping/MappingState.h [hunk 2]
```diff
@@ -152,6 +153,11 @@ class MappingState {
   const std::map<Operation *, std::vector<MappingLoc>> &getOpToLocs() const {
     return this->op_to_locs;
   }
+  const std::unordered_map<RegisterFile *,
+                           std::unordered_map<int, Operation *>> &
+  getRegFileToOccupyOperations() const {
+    return this->reg_file_to_occupy_operations;
+  }
```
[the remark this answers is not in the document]
guosran: sry, they should present now

lib/NeuraDialect/Mapping/MappingState.cpp
```diff
@@ -30,23 +30,195 @@ MappingState::MappingState(const Architecture &arch, int II,
...
+  auto it = slot_map.find(canonical_step);
+  // Enforces the one-write-per-cluster-per-slot constraint: reject any second
+  // writer, regardless of whether it targets the same register or a different
+  // one.  This disallows both:
+  //   * ADD NORTH, SOUTH -> $0, $1  (single op writing two regs in a cluster)
+  //   * ADD -> $0  +  MOV -> $1     (two ops writing different regs in a cluster)
+  if (it != slot_map.end()) {
+    return false;
+  }
+  slot_map[canonical_step] = nullptr; // Placeholder; caller identity not needed
```
tancheng: Why add nullptr here? what does this mean?

lib/NeuraDialect/Mapping/MappingState.cpp
```diff
@@ -88,82 +260,106 @@ bool MappingState::bindMultiCycleOp(BasicResource *resource, int start_time,
...
       }
     }
-    return true;
   }
+
+  // Register-cluster constraint: rejects when a *different* register in
+  // the same RegisterFile is already occupied at a congruent time slot.
+  if (loc.resource->getKind() == ResourceKind::Register) {
+    Register *reg = static_cast<Register *>(loc.resource);
+    if (!isRegisterWriteAvailableAcrossTime(reg, loc.time_step)) {
```
tancheng: Why we only checks `write`? we should check `write` for other `write`, and check `read` for other `read`. read and write can happen together. I didn't see where your `isRegisterReadAvailableAcrossTime` is used?

lib/NeuraDialect/Mapping/MappingState.cpp
tancheng: Why? I thought the `value` should be operations? `reg_file_write_to_occupy_operations`. Why previous `nullptr` also works? Do you mean you just need the `key` exist, you don't actually care the `value`? If so, we can update `reg_file_write_to_occupy_operations` to `reg_file_write_to_occupy`.

@guosran can you plz try to explain the reasoning here? I am okay with either implementation, but we can discuss a bit more.

BTW, it seems you didn't push your commits again.

guosran:
`gemm_kernel.mlir` fails during mapping even for II=20, which is the maximum. 
The root cause appears to be routing congestion on tile#10: multiple data moves (e.g., %139, %142, %151, %172) originating from tile#10 at early time steps (t=12–13) fail to find routes back to tile#10 at later time steps (t=21–26). These are predicate values produced by icmp/not operations that need to be routed to grant_predicate consumers, but all register and link paths are already occupied by other routed values holding across long time ranges. The mapper exhausts all candidate schedule slots and route options without finding a feasible placement.

tancheng:
> %139, %142, %151, %172

are they with the same source? If so, can we merge them? Large-fan-out is indeed a problem that we didn't optimize yet.

lib/NeuraDialect/Mapping/MappingState.cpp
```diff
@@ -30,23 +32,236 @@ MappingState::MappingState(const Architecture &arch, int II,
...
+bool MappingState::isRegisterReadAvailableAcrossTime(Register *reg,
+                                                     int time_step) const {
+  RegisterFile *reg_file = reg->getRegisterFile();
+  if (!reg_file) {
+    return true;
+  }
+
+  // Checks whether the cluster already has a reader at any congruent time slot
+  // targeting a different register.  Same-register reads are allowed (shared).
+  auto checkSlot = [this, reg_file, reg](int t) -> bool {
```
ShangkunLi: `checkSlot` -> `check_slot`

lib/NeuraDialect/Mapping/MappingState.cpp
guosran: Sorry I forgot to reply... yes...

PR #295, opened 2026-03-20
Enhance the logic of `phi` -> `phi_start` transform

ShangkunLi:
In this pr,
- We enhance the logic of transforming `phi` to `phi_start`. In the previous logic, we transform all `phi` with a `reserve` op as an operand into `phi_start`. This may violate the semantics in the simulator, so we only transform `phi` ops with a `reserve` op + a `grant_once` op into `phi_start` ops
- We modify the codegen pass for the corresponding changes

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,7 @@ env:
   LLVM_COMMIT: "6146a88f60492b520a36f8f8f3231e15f3cc6082"
   CCACHE_DIR: "${{ github.workspace }}/.ccache"
   CCACHE_COMPRESS: "true"
-  CCACHE_MAXSIZE: "1G"
+  CCACHE_MAXSIZE: "4G"
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


tancheng:
> > Okay, so it sounds like, if the LHS is `neura.grant_predicate`, we won't transform it to `phi_start`; if LHS is `neura.grant_once`, we do the transformation.
> > Above summary is correct?
> 
> Yes, this is what this pr done.
> 
> But after carefully thinking, I think we don't even need a `phi_start`. Because if we encounter:
> 
> 1. `phi (<val, 1>, <N/A, N/A>)` or `phi (<N/A, N/A>, <val, 1>)`: we should simply choose the one with predicate `1`
> 2. `phi (<N/A>, <N/A>)`: we should output `<AnyVal, 0>`
> 
> The semantics of `phi` is robust enough, I think.

Well, let's keep `phi_start` for now. We need `phi_start` in RTL to enable context-switch across kernels.

ShangkunLi:
> @ShangkunLi do we still want to merge this PR?

If we choose to keep `phi_start`, we don't need to check in this pr.

But the problem is that we may have some semantic overlap between `phi_start` and `phi`.

PR #310, opened 2026-04-15
Support Symbol Dynamic Task/Kernel Lowering

ShangkunLi:
## Background
Inspired by https://dl.acm.org/doi/10.1145/3779212.3790229, we can categorize the kernels/tasks into the following three categories:
- Static Loop Bound: all bounds are statically known [Already supported by counters]
- Symbol Dynamic Loop Bound: bounds are dynamic but known before this kernel/task launch [Not yet supported]
- Data-Dependent Loop Bound: bounds are computed at runtime [Already supported by NEURA compiler]

Hence, we want to support such symbol dynamic bounds, which are widely shown in many applications like transformer, MoE, conv, etc.

And supporting such symbol dynamic bounds is essential for runtime features.

## PR Summary
**Compiler / Lowering**
- Add linalg-to-affine lowering pass pipeline as a new frontend stage (`tosa` dialect does not support symbol dynamic)
- Support dynamic loop bounds in lowering passes (neura.counter now accepts runtime values)
- Add constant folding for `neura.counter` bounds
- Fix bug in copy-to-affine conversion
- Handle `expand_shape` op in the lowering pipeline
- Add math dialect support in `mlir-neura-opt`

**Task Fusion**
- Fix two bugs in `FuseTaskPass`: null-deref in `counterChainsMatch` (bounds are SSA operands, not integer attributes) and "operation destroyed but still has uses" caused by incorrect clone ordering of counter chains vs. arith constants

**Tests**
- Add end-to-end tests for Conv1D, GCN (dynamic), cross-attention, MLP, and transformer models
- Update existing tests to match new counter operand definition
- Add Python source scripts for model IR generation

**CI**
- Update `main.yml` to use `torch==2.5.1+cpu` and `torch-mlir==20251210.657` (matching the `torch-ir-sync` environment), with torch-mlir wheel self-hosted at `ShangkunLi/python-wheels` (for package stability reason)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -147,17 +147,13 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
-    # Installs NumPy 1.x first (compatable with torch-mlir)
-    - name: Install NumPy 1.x
-      run: pip install "numpy<2.0"
-    - name: Install PyTorch
+    - name: Install PyTorch (cpu)
       run: |
-        wget https://github.com/llvm/torch-mlir/releases/download/snapshot-20231229.1067/torch-2.2.0.dev20231204+cpu-cp311-cp311-linux_x86_64.whl
-        pip install torch-2.2.0.dev20231204+cpu-cp311-cp311-linux_x86_64.whl
+        pip install torch==2.5.1+cpu --index-url https://download.pytorch.org/whl/cpu
     - name: Install torch-mlir
       run: |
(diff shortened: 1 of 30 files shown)


tancheng:
I need some brief presentation/discussion/explanation about this PR..

ShangkunLi:
> I need some brief presentation/discussion/explanation about this PR..

Sure~ Do you mind if I schedule a meeting tomorrow morning (e.g., 9:30 HK Time, 18:30-19:00 LA Time)? Or other time slots when you are available?

tancheng (approved):

ShangkunLi:
Hi~ @tancheng, can we merge this pr now, so that @guosran can proceed with spatial-temporal task allocation? Or merge this after this week's discussion?

PR #319, opened 2026-06-05
Refactor Task/Counter Classification for Runtime-Managed Tasks

ShangkunLi:
In this PR:

- Split ClassifyTaskAndCounterPass into two phases: counter classification first, runtime-managed task identification second.
- Replaced old symbol_dynamic semantics with regular_dynamic counter dynamism.
- Mark tasks as runtime_managed only when they contain regular-dynamic counters and no loop-carried dependence pattern.
- Updated pass documentation and related FileCheck expectations.


include/TaskflowDialect/TaskflowPasses.td
```diff
@@ -61,16 +60,17 @@ def ClassifyTaskAndCounter : Pass<"classify-task-and-counter", "ModuleOp"> {
         "relay"  : middle loop   (has parent and children)
         "leaf"   : innermost loop (no children; also single-loop tasks)
 
-      counter_dynamism – bound predictability:
+      counter_dynamism – AMOEBA drive-pattern class:
         "static"            : all bounds are compile-time constants
-        "symbol_dynamic"    : bounds trace to function args / memref.dim
-        "irregular_dynamic" : bounds depend on earlier task outputs
+        "regular_dynamic"   : bounds are known before the task launches
```
tancheng: can this `runtime_aware`?

include/TaskflowDialect/TaskflowPasses.td
```diff
@@ -61,16 +60,17 @@ def ClassifyTaskAndCounter : Pass<"classify-task-and-counter", "ModuleOp"> {
...
+        "irregular_dynamic" : bounds are produced during task execution
```
tancheng: call this `dynamic` then?

include/TaskflowDialect/TaskflowPasses.td
ShangkunLi: Here, we classify the DFG into two categories, the regular DFG and the irregular DFG. Regular DFG means the DFG can be driven by structured counters, while irregular DFG cannot.

Among regular DFGs, we further classify them into two types based on whether their bounds are known before launch. If we know the bounds before the DFG launches, it is `static`. Otherwise, it is `regular-dynamic`.

If renaming to `runtime_aware` or `dynamic`, it would be too general. WDYT?

tancheng: then
 - static_bound
 - runtime_aware_bound
 - dynamic_bound?

ShangkunLi: How about naming them:
- static-parameterized
- runtime-parameterized
- irregular

tancheng: "irregular" sounds too general, the other two sound good~!

ShangkunLi: But in our case, all the other cases besides the first two are lowered through the `neura` flow.

It seems to be okay to make it general?

tancheng: what about constant-bound; symbol-bound; and dynamic-bound? I don't understand why we need to classify the DFG into categories, and the "regular" and "irregular" are too vague, "irregular" sometimes means ctrl-intensive DFGs, has limited overlap with the "bound".

ShangkunLi: Updated accordingly~

include/TaskflowDialect/TaskflowPasses.td
```diff
@@ -61,16 +60,21 @@ def ClassifyTaskAndCounter : Pass<"classify-task-and-counter", "ModuleOp"> {
...
+        "constant_bound" : all bounds are compile-time constants
+        "symbol_bound"   : bounds are known before the task launches
+        "dynamic_bound"  : bounds are produced during task execution
 
       counter_id – unique integer index within the task (0-based)
 
     Per task:
-      task_type – worst-case counter_dynamism across all counters
-        ("irregular_dynamic" > "symbol_dynamic" > "static")
+      dlp_eligibility = "replicable" when the task has hyperblocks without
```
tancheng: How about `dlp_replicable = true`?

include/TaskflowDialect/TaskflowPasses.td
```diff
@@ -61,16 +60,21 @@ def ClassifyTaskAndCounter : Pass<"classify-task-and-counter", "ModuleOp"> {
...
+      loop-carried dependence and can be considered for data-level parallel
+      replication.
+
+      task_type = "runtime_managed" only when the task has at least one
```
tancheng: Is there any other task type, besides this `runtime_managed`?

ShangkunLi: For now, I do not figure out other types of tasks that are needed. This is hardware-specific. We need more hardware support for more task types.

include/TaskflowDialect/TaskflowPasses.td
ShangkunLi: Updated~

include/TaskflowDialect/TaskflowPasses.td
tancheng: Then it sounds like `runtime_managable = true`?

ShangkunLi: Updated accordingly~

tancheng (approved):

PR #323, opened 2026-06-10
Fix crash on non-empty return_void blocks

ShangkunLi:
The --canonicalize-return pass previously assumed that every block containing a neura.return_void operation was an empty block (i.e., return_void was the only operation in it). When this assumption was violated — which happens for real-world kernels where other operations share the same basic block as return_void — the pass hit an unconditional assert(false) and crashed:
```c++
assert(false && "Unsupported case: return block is not empty.");
```

This made the pass fail on a broad class of benchmarks (fir, conv, spmv, jacobi, etc.) that produce non-empty return_void blocks after lowering.

Fix:
- Removed the `is_empty_block` guard and the failing assert in both call sites inside `processVoidReturnsInKernel` and `CanonicalizeReturnPass`.
- Renamed `processEmptyReturnVoidBlock` → `processReturnVoidBlock` to reflect that it now handles the general case.
- The underlying predecessor-walking logic already handled non-empty blocks correctly; the guard was overly conservative.

### DIFF
--- a/lib/NeuraDialect/Transforms/CanonicalizeReturnPass.cpp
+++ b/lib/NeuraDialect/Transforms/CanonicalizeReturnPass.cpp
@@ -132,9 +132,9 @@ static void processYields(neura::KernelOp kernel_op, OpBuilder &builder) {
   }
 }
 
-static void processEmptyReturnVoidBlock(Block *ret_block,
-                                        neura::ReturnOp void_ret_op,
-                                        OpBuilder &builder) {
+static void processReturnVoidBlock(Block *ret_block,
+                                   neura::ReturnOp void_ret_op,
+                                   OpBuilder &builder) {
   SmallVector<Block *> predecessor_blocks(ret_block->getPredecessors());
   // Entry bolock with return_void is unreachable; no action needed.
   if (predecessor_blocks.empty()) {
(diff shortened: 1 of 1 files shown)


lib/NeuraDialect/Transforms/CanonicalizeReturnPass.cpp
```diff
@@ -337,18 +332,7 @@ struct CanonicalizeReturnPass
...
-        // Checks if ret_block only contains the return_void operation.
-        bool is_empty_block = (ret_block->getOperations().size() == 1);
-
-        if (is_empty_block) {
-          processEmptyReturnVoidBlock(ret_block, ret_void_op, builder);
-        } else {
-          // TODO: Handle non-empty return blocks.
-          // The basic idea is to create a new block that only contains the
-          // return_void operation, and redirect the original return block to
-          // this new block.
```
tancheng, line 349: Why we previously want a standalone bb only containing `return`? Can you recall?

tancheng (commented):
Why we don't need to update any tests? They all already passed?

tancheng:
@ShangkunLi sorry, I forgot how do we want proceed?

ShangkunLi:
> @ShangkunLi sorry, I forgot how do we want proceed?

As we discussed last week, we may first merge this pr, and then refactor the code based #325.

tancheng (approved):

PR #330, opened 2026-06-25
Refine Taskflow Memory Dependency Semantics

ShangkunLi:
This PR:

1. **Fix false read-read serialization in `AffineToTaskflow`.**

   Previously, the conversion used a single memref value mapping to represent memory dependence states. This could incorrectly serialize two tasks that only read the same memref.

   Conceptually:

   ```text
   T0: read A -> write B
   T1: read A -> write C
   ```

   These two tasks should both read the same current state of `A` and run independently. With the old mapping, `T1` could accidentally consume a read state produced by `T0`, creating a fake edge.

   The pass now separates memory state tracking into latest write states and pending read states. Independent readers consume the same latest write state and do not depend on each other.

2. **Emit `dependency_read_out` only for real WAR dependencies.**

   Previously, every task yielded all read memrefs as `dependency_read_out`, even if no later writer needed that read state.

   Old IR example from `multi-nested.mlir`:

   ```mlir
   %dependency_read_out, %dependency_write_out =
     taskflow.task @Task_0
       dependency_read_in(%arg0 : memref<?x8x6xi32>)
       dependency_write_in(%arg5 : memref<?xi32>)
       ...
       : (memref<?x8x6xi32>, memref<?xi32>)
      -> (memref<?x8x6xi32>, memref<?xi32>) {
     ...
     taskflow.yield reads(%arg10 : memref<?x8x6xi32>)
                    writes(%arg11 : memref<?xi32>)
   }
   ```

   New IR:

   ```mlir
   %dependency_write_out =
     taskflow.task @Task_0
       dependency_read_in(%arg0 : memref<?x8x6xi32>)
       dependency_write_in(%arg5 : memref<?xi32>)
       ...
       : (memref<?x8x6xi32>, memref<?xi32>)
      -> (memref<?xi32>) {
     ...
     taskflow.yield writes(%arg11 : memref<?xi32>)
   }
   ```

   For true WAR cases, the pass still emits sparse read states. For example, in `cross_attention.mlir`, `Task_3` only yields the read state that a later writer must wait on:

   ```mlir
   taskflow.yield reads(%arg3 : memref<64x64xf32>)
                  writes(%arg5 : memref<?x64xf32>)
   ```

   This keeps the IR cleaner while still preserving RAW/WAW/WAR ordering correctly.

The PR also updates `taskflow.task` documentation, parser/printer behavior, and tests to match the new dependence-state semantics.

### DIFF
--- a/include/TaskflowDialect/TaskflowOps.td
+++ b/include/TaskflowDialect/TaskflowOps.td
@@ -31,23 +31,27 @@ def TaskflowTaskOp
   let summary = "Computation task operation within a Taskflow graph.";
 
   let description = [{
-    Represents a computational task that takes data inputs and produces
-    data outoputs. Tasks are isolated from their surrounding scope and can only
-    communicate through explicit data dependencies.
+    Represents a computational task that consumes and produces explicit
+    dependence-state SSA values. Tasks are isolated from their surrounding scope
+    and can only communicate through these explicit dependencies.
 
     Tasks has two types of inputs/outputs:
-    1. Memory dependencies: memrefs that are read or written by the task
+    1. Memory dependence states: memref SSA versions that the task reads or
+       writes. A dependency_read_out or dependency_write_out is still the same
+       underlying memref, but represents the state after this task has observed
+       or modified it.
     2. Value dependencies: SSA values from producer tasks
 
-    The `original_read_memrefs` and `original_write_memrefs` attributes record the actural
-    original memrefs that this task accesses,
-    enabling data placement analysis for multi-CGRA mapping.
+    The `original_read_memrefs` and `original_write_memrefs` operands record
+    the actual original memrefs that this task accesses. These operands keep
+    the memory identity separate from the dependence-state SSA value, enabling
+    dependence analysis and data placement for multi-CGRA mapping.
 
     Example:
-      // Memory inputs: %mem, Value inputs: %val
+      // Memory state inputs: %mem_state, %out_state. Value input: %val.
       $out_mem, %out_val = taskflow.task "Task_0"
-        dependency_read_in(%mem : memref<4xi32>)
-        dependency_write_in(%arg5 : memref<?xi32>)
+        dependency_read_in(%mem_state : memref<4xi32>)
+        dependency_write_in(%out_state : memref<?xi32>)
         value_inputs(%val : i32)
         [original_read_memrefs(%arg0 : memref<?x8x6xi32>),
          original_write_memrefs(%arg5 : memref<?xi32>)] {
@@ -82,8 +86,10 @@ def TaskflowYieldOp
                                ParentOneOf<["TaskflowTaskOp"]>]> {
   let summary = "Yield operation for Taskflow task";
   let description = [{
-    Yields values from a task body. The number and types of operands
-    must match the result types of the parent taskflow.task operation.
+    Yields dependence-state and value results from a task body. Read results
+    are sparse: a task only needs to yield a read state when a later writer must
+    depend on that read for WAR ordering. The number and types of operands must
+    match the result types of the parent taskflow.task operation.
     
     Example:
       taskflow.task "Task_0" (%arg0, %arg1) {
--- a/lib/Conversion/AffineToTaskflow/AffineToTaskflowPass.cpp
+++ b/lib/Conversion/AffineToTaskflow/AffineToTaskflowPass.cpp
@@ -91,12 +91,22 @@ static void collectExternalValues(Operation *root_op,
   }
 }
 
-// Updates operands of an operation using the value mapping.
+// Updates operands of a non-loop operation after preceding loops have been
+// converted into taskflow.task ops. Memref operands consume the latest write
+// state for that original memref; scalar SSA values use value_mapping.
 static void
 updateOperationOperands(Operation *op,
+                        const DenseMap<Value, Value> &latest_write_out,
                         const DenseMap<Value, Value> &value_mapping) {
   for (OpOperand &operand : op->getOpOperands()) {
     Value original_value = operand.get();
+    if (isa<MemRefType>(original_value.getType())) {
+      auto it = latest_write_out.find(original_value);
+      if (it != latest_write_out.end())
+        operand.set(it->second);
+      continue;
+    }
+
     auto it = value_mapping.find(original_value);
     if (it != value_mapping.end()) {
       operand.set(it->second);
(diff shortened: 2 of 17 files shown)


tancheng (approved):
Can you remind me the diff between `dependency_read_out` and `dependency_read_in`?

ShangkunLi:
> Can you remind me the diff between `dependency_read_out` and `dependency_read_in`?

`dependency_read_in` is the memory state a task must observe before it can safely read a memref.

`dependency_read_out` is the memory state after the task has performed that read. It is only needed when a later writer must wait for this read to finish, i.e. for WAR ordering.

Simple example:

```text
T0: read A
T1: write A
```

IR shape:

```mlir
%a_after_read = taskflow.task @T0
  dependency_read_in(%A : memref<...>)
  [original_read_memrefs(%A : memref<...>)]
  : (memref<...>) -> (memref<...>) {
  ...
  taskflow.yield reads(%A_arg : memref<...>)
}

%a_after_write = taskflow.task @T1
  dependency_write_in(%a_after_read : memref<...>)
  [original_write_memrefs(%A : memref<...>)]
  : (memref<...>) -> (memref<...>) {
  ...
  taskflow.yield writes(%A_arg : memref<...>)
}
```

tancheng:
> > Can you remind me the diff between `dependency_read_out` and `dependency_read_in`?
> 
> `dependency_read_in` is the memory state a task must observe before it can safely read a memref.
> 
> `dependency_read_out` is the memory state after the task has performed that read. It is only needed when a later writer must wait for this read to finish, i.e. for WAR ordering.
> 
> Simple example:
> 
> ```
> T0: read A
> T1: write A
> ```
> 
> IR shape:
> 
> ```mlir
> %a_after_read = taskflow.task @T0
>   dependency_read_in(%A : memref<...>)
>   [original_read_memrefs(%A : memref<...>)]
>   : (memref<...>) -> (memref<...>) {
>   ...
>   taskflow.yield reads(%A_arg : memref<...>)
> }
> 
> %a_after_write = taskflow.task @T1
>   dependency_write_in(%a_after_read : memref<...>)
>   [original_write_memrefs(%A : memref<...>)]
>   : (memref<...>) -> (memref<...>) {
>   ...
>   taskflow.yield writes(%A_arg : memref<...>)
> }
> ```

Can you try to propose some new names... I feel it is a bit hard to understand..

ShangkunLi:
> > > Can you remind me the diff between `dependency_read_out` and `dependency_read_in`?
> > 
> > 
> > `dependency_read_in` is the memory state a task must observe before it can safely read a memref.
> > `dependency_read_out` is the memory state after the task has performed that read. It is only needed when a later writer must wait for this read to finish, i.e. for WAR ordering.
> > Simple example:
> > ```
> > T0: read A
> > T1: write A
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > IR shape:
> > ```mlir
> > %a_after_read = taskflow.task @T0
> >   dependency_read_in(%A : memref<...>)
> >   [original_read_memrefs(%A : memref<...>)]
> >   : (memref<...>) -> (memref<...>) {
> >   ...
> >   taskflow.yield reads(%A_arg : memref<...>)
> > }
> > 
> > %a_after_write = taskflow.task @T1
> >   dependency_write_in(%a_after_read : memref<...>)
> >   [original_write_memrefs(%A : memref<...>)]
> >   : (memref<...>) -> (memref<...>) {
> >   ...
> >   taskflow.yield writes(%A_arg : memref<...>)
> > }
> > ```
> 
> Can you try to propose some new names... I feel it is a bit hard to understand..、

Maybe we can name them after their role as ordering states instead of just in/out:

- `dependency_read_in` -> `read_precondition`
- `dependency_read_out` -> `read_completion`
- `dependency_write_in` -> `write_precondition`
- `dependency_write_out` -> `write_completion`

The “completion” names make it clearer that these are not loaded/stored data values; they are ordering states produced after the read/write has completed.

Or:

- `dependency_read_in` -> `read_waits`
- `dependency_read_out` -> `read_dones`
- `dependency_write_in` -> `write_waits`
- `dependency_write_out` -> `write_dones`

The idea is:
- `read_waits`: states this task must wait on before reading
- `read_dones`: states produced after this task's reads complete, used by later writers for WAR
- `write_waits`: states this task must wait on before writing
- `write_dones`: states produced after this task's writes complete, used by later readers/writers for RAW/WAW

tancheng:
What about
 - will_read
 - will_write
 - done_read
 - done_write

tancheng:
And I now feel `tensor` (SSA) is more suitable for taskflow. `memref` is after bufferization, which is too low level.. It would be perfect if taskflow can take both `tensor` and `memref` as operands..

ShangkunLi:
> And I now feel `tensor` (SSA) is more suitable for taskflow. `memref` is after bufferization, which is too low level.. It would be perfect if taskflow can take both `tensor` and `memref` as operands..

Ok, I will try to figure out whether we can do that.

ShangkunLi:
> What about
> 
> * will_read
> * will_write
> * done_read
> * done_write

Sounds good to me, I will modify this only in the next pr.

ShangkunLi:
> > And I now feel `tensor` (SSA) is more suitable for taskflow. `memref` is after bufferization, which is too low level.. It would be perfect if taskflow can take both `tensor` and `memref` as operands..
> 
> Ok, I will try to figure out whether we can do that.

I think the clean design is to let Taskflow support both tensor SSA dataflow and memref memory-state dependencies, but keep their semantics separate.

For `tensor` values, Taskflow can use normal data/value inputs and outputs, because SSA already represents producer-consumer dependence. There is no WAR/WAW issue for tensors.

For `memrefs`, we still need explicit memory-state dependencies like `will_read`, `will_write`, `done_read`, and `done_write`, because memrefs are mutable and may alias.

So the proposal is:
- support tensor/scalar values through task data/value inputs and outputs;
- keep memref read/write states only for bufferized memory effects;
- add a tensor-level taskflow conversion pass before bufferization, and lower/bufferize it later into the existing memref-level taskflow/neura path.

tancheng:
> > > And I now feel `tensor` (SSA) is more suitable for taskflow. `memref` is after bufferization, which is too low level.. It would be perfect if taskflow can take both `tensor` and `memref` as operands..
> > 
> > 
> > Ok, I will try to figure out whether we can do that.
> 
> I think the clean design is to let Taskflow support both tensor SSA dataflow and memref memory-state dependencies, but keep their semantics separate.
> 
> For `tensor` values, Taskflow can use normal data/value inputs and outputs, because SSA already represents producer-consumer dependence. There is no WAR/WAW issue for tensors.
> 
> For `memrefs`, we still need explicit memory-state dependencies like `will_read`, `will_write`, `done_read`, and `done_write`, because memrefs are mutable and may alias.
> 
> So the proposal is:
> 
> * support tensor/scalar values through task data/value inputs and outputs;
> * keep memref read/write states only for bufferized memory effects;
> * add a tensor-level taskflow conversion pass before bufferization, and lower/bufferize it later into the existing memref-level taskflow/neura path.

Sounds good to me. I think we can have the taskflow supporting tensor, then implement bufferization interface for it. So the bufferization pass serves as the lowering from `tensor` to `memref`.

PR #332, opened 2026-06-26
Rename Taskflow memory dependency

ShangkunLi:
Rename Taskflow memory dependency states from `dependency_*_in/out` to `will_read`, `will_write`, `done_read`, and `done_write` across op assembly, accessors, implementation, and tests.


tancheng (approved):
Sounds like it should be `reads` instead `read`, `writes` instead of `write`?

ShangkunLi:
> Sounds like it should be `reads` instead `read`, `writes` instead of `write`?

Updated~

PR #273, opened 2026-02-23
Add Python frontend for Neura hardware custom ops

YanzhouTang:
## Summary

- Introduces `python/neura_ops.py` as the Python frontend interface for Neura hardware primitives.
- Programmers can explicitly call custom ops (e.g., `torch.ops.neura.gather()`) to annotate hardware-specific memory access patterns that standard PyTorch ops cannot express.
- Custom ops behave identically to standard PyTorch operations at runtime, but are preserved as-is through torch_mlir tracing so that the Neura compiler backend can recognize and lower them to hardware primitives.

## Example: `neura::gather` in hash encoding

- `hash_encode` kernel uses indirect addressing (`level_embeddings[indices]`) for hash-table lookups. After torch_mlir tracing, this becomes `aten.index.Tensor`, which loses the gather semantics needed by CGRA.
- With the custom op, programmers write `torch.ops.neura.gather(table, indices)` instead. torch_mlir preserves it as `torch.operator "neura.gather"`, giving the compiler a clear signal.
- Refactored `trilinear_interpolation` to batch all 8 corner indices into a single gather per level (from 8 calls down to 1).

## Changes

1. **`python/neura_ops.py`** (new): Registers `neura::gather` with CPU and Meta implementations, compatible with PyTorch 2.1+.
2. **`nerf_kernels.py`** (new): Imports `neura_ops`, replaces fancy indexing with `torch.ops.neura.gather()`, two-phase trilinear interpolation.
3. **`compile_hash_encode.py`** (new): Compilation script that exports clean ~930-line Torch Dialect MLIR with `neura.gather` preserved.

## Verification

- `neura.gather` appears exactly 2 times in output MLIR (one per level).
- `aten.index.Tensor` fully eliminated.
- PyTorch-side correctness verified (semantically equivalent to the original implementation).

PR #339, opened 2026-07-16
Add neura.gather op and torch-to-neura lowering pass

YanzhouTang:
## Summary

Follow-up to #273. That PR defined the PyTorch custom op `neura::gather` on the
frontend side. This PR implements lowering it from PyTorch down to Neura, and
adds a unit test.

## Changes

- Defines the native `neura.gather` dialect op (`NeuraOps.td`).
- Adds the `lower-torch-to-neura` pass that rewrites the frontend marker
  `torch.neura.gather` into `neura.gather` (`lib/Conversion/TorchToNeura/`).
- Adds a unit test (`test/frontend/gather/`) that drives the op through the
  standard lowering pipeline and checks it becomes `neura.gather`.


test/frontend/gather/gather.mlir
```diff
@@ -1,12 +1,24 @@
...
+// the marker becomes a native neura.gather op and no torch marker remains.
+
+// RUN: python3 %S/gather.py %t.torch.mlir
+// RUN: FileCheck --input-file=%t.torch.mlir %s --check-prefix=L1 --implicit-check-not="aten.index"
+
+// L1-LABEL: func.func @forward
+// L1: torch.operator "torch.neura.gather"
+
+// RUN: python3 %S/gather.py --neutral %t.neutral.mlir
+// RUN: mlir-neura-opt -allow-unregistered-dialect -lower-torch-to-neura %t.neutral.mlir | FileCheck %s --check-prefix=L2 --implicit-check-not="torch.neura.gather"
```
ShangkunLi, line 21: IIRC, there is `torch` dialect https://github.com/llvm/torch-mlir. Shall we include it in our tool so that we can avoid adding this unsafe `--allow-unregistered-dialect` pass here?

YanzhouTang: Makes sense to put it in the Torch dialect long-term. But that means rebuilding torch-mlir and publishing a new wheel. The current wheel only ships the Python runtime, so we can't link the dialect into our tool yet. Hence the temporary `--allow-unregistered-dialect` bypass here.

Since our CI already uses the wheel from your ShangkunLi/python-wheels repo, is there a precedent for adding a custom op into torch-mlir that we can follow?

test/frontend/gather/compile_gather.py
tancheng, line 1: Shouldn't we rename this file based on its usage?
e.g., `compile_and_test_gather.py`

tancheng (approved):

test/frontend/gather/compile_gather.py
YanzhouTang: Renamed `gather.py` to `compile_gather.py`

test/frontend/gather/gather.mlir
tancheng: @ShangkunLi will investigate how hard to include torch ir as submodule.

PR #340, opened 2026-07-27
Analytical cost model + exact mapping (step 1: single-CGRA)

guosran:
## Motivation

The long-term goal is to map **whole real programs** onto multi-CGRA fabrics. Two problems block that today:

1. **Multi-task mapping ignores inter-task communication.** When several tasks are co-mapped, the cost of moving data *between* CGRAs is not modelled, so the global mapping is not actually optimized for the thing that dominates on real programs.
2. **The search explodes as programs get complex.** Two compounding blow-ups:
   - *Per-shape:* choosing the CGRA shape for a single task currently re-runs the **entire lowering + mapping pipeline for every candidate shape**. That is combinatorial and does not scale.
   - *Per-mapping:* the mapper itself is a **greedy heuristic** with backtracking. On non-trivial kernels it is **sub-optimal** (leaves II on the table) and its runtime **explodes** — and on some kernels it fails to find any mapping at all.

The plan is to replace the "run the full pipeline everywhere" approach with a **cheap, accurate analytical cost model** that ranks shapes/mappings in O(1), and to run the expensive exact mapper only where it matters — cutting time *and* improving quality.

**This PR is step 1: use the cost model to optimize single-CGRA, single-task mapping, and fix the shortcomings we found in the existing mapper.** Next step: single-task **multi-CGRA**. Ultimate goal: the whole real program, including inter-task communication.

### 1. Analytical cost model (O(1) II prediction)
`--cost-model-analytical` predicts the modulo-scheduling II directly from the DFG + architecture, with no mapper run:

```
II = max(ResMII, RecMII, MemMII, RouteMII, RegMII, IssueMII)
```

Each bound is an interpretable `ceil(demand / capacity)` term (recurrence, per-FU-class resource, memory ports, routing bandwidth, register pressure, issue width). No training, no lookup tables, no workload names. This is the cheap ranking function that replaces running the full pipeline per candidate shape.

### 2. Exact CP-SAT mapper as ground truth + production mapper
`test/cost-model/exact_mapper_cpsat.py` models the **full** mapping — placement + modulo schedule + the concrete routing of every value over the real link graph, under modulo resource limits — and returns a provably valid mapping at the true contention-aware optimum. It is used both to validate the cost model and, via the bridge below, as the mapper of record where the heuristic is inadequate.

### 3. Bridge the exact mapping back into the compiler (`--import-mapping`)
The exact mapper emits its concrete placement **and every value's routing path**; `--map-to-accelerator="import-mapping=<json>"` replays that exact solution (bind each op, reserve each route) so the existing `--generate-code` emits real YAML/ASM. This is what turns "an optimal schedule exists" into "the compiler produces the optimal binary".

## Shortcomings of the previous mapper — and the fixes

| Problem in main's heuristic mapper | Fix in this PR | Result (vs. main) |
|---|---|---|
| **Sub-optimal II** — greedy backtracking settles above the true optimum | Cost model flags the gap; exact mapper finds a smaller, routable II; `import-mapping` adopts it | histogram **6 → 5**, bicg **10 → 9**, spmv **15 → 10**, fft **20 → 19** |
| **Search cost grows steeply** on larger kernels (deep backtracking) | Exact solve run once on the chosen shape, with an `II+1` fallback so a routable mapping is always produced within budget | all seven kernels map; the exact solve replaces an unbounded heuristic search |
| **Greedy per-net routing** can't reproduce a joint routing on large kernels | Emit and **replay the exact routes** instead of re-routing greedily | spmv (139 ops) / fft (150 ops) import cleanly; deterministic |

## Results (single 4×4 CGRA)

| kernel | prev heuristic II | adopted II | Δ |
|---|---|---|---|
| histogram | 6 | **5** | −1 |
| bicg | 10 | **9** | −1 |
| spmv | 15 | **10** | −5 |
| fft | 20 | **19** | −1 |
| fir | 5 | 5 | 0 (already optimal) |
| relu | 5 | 5 | 0 (already optimal) |
| gemm | 17 | 17 | 0 |

Honest boundaries: for **gemm** the cost-model lower bound (13) is *not* tight — the exact solver proves II 13/14 are infeasible, so the true optimum is ≥15 and we keep main's 17. For **spmv/fft** the adopted II is the smallest the solver reaches within the time budget (spmv's true optimum is 9; more budget lowers it further), but both are strictly below main's heuristic result (15 and 20).

### Analytical Lower Bound vs. compiled II — next step

| kernel | analytical LB | compiled II | gap |
|---|---|---|---|
| fir | 5 | 5 | 0 |
| relu | 5 | 5 | 0 |
| histogram | 5 | 5 | 0 |
| bicg | 9 | 9 | 0 |
| spmv | 9 | 10 | 1 |
| fft | 16 | 19 | 3 |
| gemm | 13 | 17 | 4 |

The gap is placement-dependent routing contention, invisible to a structural bound. **Next:** turn the lower bound into a calibrated *predictor* of the achievable II (still no training/lookup/workload names) so it lands near compiled II — needed once multi-CGRA / inter-task makes per-candidate exact verification too costly.

### Timing (vs. main's flow)
- **Cost model:** ~ms per shape. This is the piece that scales: it replaces main's approach of re-running the **full lowering + mapping pipeline for every candidate shape**, which is the combinatorial blow-up we are removing.
- **Exact mapper, run once on the chosen shape:** ~1–2 s (small kernels) to tens of seconds (large kernels); it scales with kernel size, and more solver budget lowers the II further on the hardest kernels. It is the deliberately-expensive step that the cheap cost model lets us run **only where it matters**, instead of everywhere.
- **Import + generate-code:** 0.1–2 s, deterministic.

Implementation note (not a result): the emit stops as soon as it has a compact, routable witness rather than proving the makespan optimal, and uses an `II+1` fallback, which keeps the per-shape solve to the times above.

## Tests
The single-CGRA e2e tests (`test/e2e/{fir,relu,histogram,bicg,spmv,gemm,fft}`) now run the unified flow end-to-end: cost-model prediction (`CM` check) → adopt the exact mapping via `import-mapping` → `generate-code`.
## Roadmap
- **This PR:** single-CGRA, single-task — cost model + exact mapping + heuristic-mapper fixes. ✅
- **Next:** single-task **multi-CGRA** (shape selection across regular/irregular multi-CGRA layouts, ranked by the cost model).
- **Ultimate:** whole real program — model **inter-task communication** so the analytical model can drive a globally optimal all-task mapping.


guosran:
Still kind of a premature idea, we may discuss more :)

PR #342, opened 2026-08-23
Add expand-memref-copy pass to unblock linalg.matmul -> Neura ingestion

tancheng:
## Summary
- `-buffer-results-to-out-params` (standard bufferization for a tensor-returning function) leaves a trailing `memref.copy` that no existing Neura lowering pass understands, which breaks the `linalg` -> `affine` -> Neura path for any kernel shaped like what torch-mlir produces for real models (compute into a temp buffer, copy into the out-param).
- Adds `--expand-memref-copy`, a small new pass that rewrites `memref.copy(src, dst)` into an explicit `affine.for` loop with `affine.load`/`affine.store`. This lets it flow through the already-supported `--lower-affine-to-neura` path like the rest of a kernel body, instead of needing dedicated Neura-side support for `memref.copy` itself. Tried `-drop-equivalent-buffer-results` and `-memref-expand` first; neither touches this case.
- Confirms (and documents via the new e2e test) that the existing `--lower-affine-to-neura` path — not a detour through LLVM dialect — is the right route for structured loop-nest kernels (matmul/conv/elementwise) coming from `linalg`, matching the pass's own stated design intent of avoiding an LLVM round-trip.

## Test plan
- [x] New lit test `test/Conversion/linalg2neura/expand-memref-copy.mlir` — the pass in isolation.
- [x] New lit test `test/Conversion/linalg2neura/matmul.mlir` — a tensor-level `linalg.matmul` (bufferize -> expand-memref-copy -> `-convert-linalg-to-affine-loops` -> the full affine-sourced Neura pipeline) mapped end-to-end onto `test/arch_spec/architecture.yaml`, with `FileCheck` verifying no `linalg.matmul`/`memref.copy`/`affine.for` survive and the expected Neura ops (`loop_control`, `load_indexed`, `fmul`, `fadd`, `store_indexed`) are present.
- [x] Full existing suite still passes: 93/94 (only `frontend/gather/gather.mlir`, which needs `torch` installed, unrelated to this change).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/test/Conversion/linalg2neura/expand-memref-copy-dynamic-and-dead.mlir
+++ b/test/Conversion/linalg2neura/expand-memref-copy-dynamic-and-dead.mlir
@@ -1,0 +1,27 @@
+// RUN: mlir-neura-opt %s --expand-memref-copy | FileCheck %s
+
+// Dynamic shape: bounds must come from memref.dim, not a constant.
+func.func @copy_dynamic(%src: memref<?x4xf32>, %dst: memref<?x4xf32>) {
+  memref.copy %src, %dst : memref<?x4xf32> to memref<?x4xf32>
+  return
+}
+
+// CHECK-LABEL: func.func @copy_dynamic
+// CHECK: %[[C0:.*]] = arith.constant 0 : index
+// CHECK: %[[DIM:.*]] = memref.dim %arg0, %[[C0]]
+// CHECK: affine.for %{{.*}} = 0 to %[[DIM]]
+// CHECK: affine.for %{{.*}} = 0 to 4
+// CHECK: affine.load %arg0
+// CHECK: affine.store %{{.*}}, %arg1
+// CHECK-NOT: memref.copy
+
+// Dead copy: target has no other users, so the copy should just disappear.
+func.func @copy_dead(%src: memref<4xf32>) {
+  %dst = memref.alloc() : memref<4xf32>
+  memref.copy %src, %dst : memref<4xf32> to memref<4xf32>
+  return
+}
+
+// CHECK-LABEL: func.func @copy_dead
+// CHECK-NOT: affine.for
+// CHECK-NOT: memref.copy
--- a/test/Conversion/linalg2neura/expand-memref-copy-dynamic-and-dead.mlir
+++ b/test/Conversion/linalg2neura/expand-memref-copy-dynamic-and-dead.mlir
@@ -1,4 +1,6 @@
-// RUN: mlir-neura-opt %s --expand-memref-copy | FileCheck %s
+// RUN: mlir-neura-opt %s --expand-memref-copy \
+// RUN: -o %t-expand.mlir
+// RUN: FileCheck %s --input-file=%t-expand.mlir
 
 // Dynamic shape: bounds must come from memref.dim, not a constant.
 func.func @copy_dynamic(%src: memref<?x4xf32>, %dst: memref<?x4xf32>) {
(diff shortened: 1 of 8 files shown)


tancheng:
Hi @ShangkunLi, `amoeba` lower `memref.copy` to `affine`, why such lowering doesn't happen in neura? I am curious about the boundary between the two in terms of dialects. Can you explicitly list what `neura` should expect as input dialects/operations?

(I feel even tensor and linalg can be `neura` inputs)

ShangkunLi:
> Hi @ShangkunLi, `amoeba` lower `memref.copy` to `affine`, why such lowering doesn't happen in neura? I am curious about the boundary between the two in terms of dialects. Can you explicitly list what `neura` should expect as input dialects/operations?
> 
> (I feel even tensor and linalg can be `neura` inputs)

From a long-term perspective, the whole compilation flow is:
```
applications/models -> (tensor operator graph) -> task graph -> kernels -> intra-kernel mapping
```

In this flow, `neura` acts as the kernel compiler. It should compile an already identified kernel, which may be represented using dialects like `linalg`/`affine`/`scf`/`llvm`/`neura`. The kernel IR should be enclosed in a neura.kernel region so that Neura has an explicit compilation boundary.

And for `amoeba`, it acts as the graph-level and kernel-orchestration compiler. It is responsible for task formation, kernel fusion, inter-kernel communication, resource allocation & scheudling.

In principle, a tile-array kernel is not tied to one specific input dialect. Its body may be represented at any abstraction level for which Neura provides a legal lowering path, including `tensor`/`linalg`/`affine`/`scf`/`llvm`/`neura`.

However, in the complete compiler flow, `amoeba` may have already consumed much of the application-level information for task formation, fusion, and inter-kernel communication. Therefore, `neura` will often receive a more canonical kernel-level representation rather than the original high-level IR. This is not a strict requirement: if `tensor` or `linalg` structure is still useful for operator-template matching or other intra-kernel optimizations, it should be preserved inside the `neura.kernel` instead of being lowered prematurely.

This design is inspired by contemporary AI compiler stacks. For example, `TorchInductor` roughly follows:
```
tensor operator graph → scheduler nodes → generated kernels (using Triton)
```
In this compilation flow, `TorchInductor` is the high-level graph compiler and `Triton` is the low-level kernel compiler.

If we follow the `synapse` + `amoeba` + `neura` design, it would be extendable to existing spatial architectures (e.g., amd npu, tenstorrent, multi-cgra, and other spatial architectures). And as far as I know, other commercial spatial architectures have a bad support for the complete compilation flow.

tancheng:
Then I think this PR is required for `neura`. And `amoeba` can leverage (instead of duplicating) this pass if needed. WDYT?

ShangkunLi:
> Then I think this PR is required for `neura`. And `amoeba` can leverage (instead of duplicating) this pass if needed. WDYT?

`amoeba` already has a pass https://github.com/coredac/amoeba/blob/main/lib/Conversion/AffineToTaskflow/ConvertCopyToAffineLoopsPass.cpp. If we plan to support this conversion only for `neura` dialects, then we should add this pass.

If other backends of `amoeba` also want to use this pass, we should keep `ConvertCopyToAffineLoopsPass.cpp`.

tancheng:
> > Then I think this PR is required for `neura`. And `amoeba` can leverage (instead of duplicating) this pass if needed. WDYT?
> 
> `amoeba` already has a pass https://github.com/coredac/amoeba/blob/main/lib/Conversion/AffineToTaskflow/ConvertCopyToAffineLoopsPass.cpp. If we plan to support this conversion only for `neura` dialects, then we should add this pass.
> 
> If other backends of `amoeba` also want to use this pass, we should keep `ConvertCopyToAffineLoopsPass.cpp`.

Then let's keep two for now? Can you help LGTM the PR?

test/Conversion/linalg2neura/expand-memref-copy-dynamic-and-dead.mlir
```diff
@@ -0,0 +1,27 @@
...
+}
+
+// CHECK-LABEL: func.func @copy_dynamic
+// CHECK: %[[C0:.*]] = arith.constant 0 : index
+// CHECK: %[[DIM:.*]] = memref.dim %arg0, %[[C0]]
+// CHECK: affine.for %{{.*}} = 0 to %[[DIM]]
+// CHECK: affine.for %{{.*}} = 0 to 4
+// CHECK: affine.load %arg0
+// CHECK: affine.store %{{.*}}, %arg1
+// CHECK-NOT: memref.copy
```
ShangkunLi, line 21: Avoid using regular expressions in the check scripts

test/Conversion/linalg2neura/expand-memref-copy-dynamic-and-dead.mlir
```diff
@@ -0,0 +1,27 @@
+// RUN: mlir-neura-opt %s --expand-memref-copy | FileCheck %s
```
ShangkunLi: It would be better to use this form for checking:
```
// RUN: mlir-neura-opt %s --expand-memref-copy \
// RUN: -o %t-expand.mlir
// RUN: FileCheck --input-file=%t-expand.mlir
```

Because using `// RUN: mlir-neura-opt %s --expand-memref-copy | FileCheck %s` may mix the debug outputs and IR outputs.

test/Conversion/linalg2neura/expand-memref-copy.mlir
```diff
@@ -0,0 +1,13 @@
...
+
+func.func @copy_2d(%src: memref<4x4xf32>, %dst: memref<4x4xf32>) {
+  memref.copy %src, %dst : memref<4x4xf32> to memref<4x4xf32>
+  return
+}
+
+// CHECK-LABEL: func.func @copy_2d
+// CHECK: affine.for
+// CHECK: affine.for
+// CHECK: %[[V:.*]] = affine.load %arg0
```
ShangkunLi: Same problem

test/Conversion/linalg2neura/expand-memref-copy-dynamic-and-dead.mlir
tancheng: Tightened the CHECK lines to exact literal matches instead of `.*`/`{{...}}` wildcards — the pass's output for a fixed input is fully deterministic, so there's no need for regex here. Fixed in 557ed13.

test/Conversion/linalg2neura/expand-memref-copy-dynamic-and-dead.mlir
tancheng: Switched to writing the pass output to a file and running FileCheck against that file, matching your suggested form. Applied the same fix to all three tests in this PR. Fixed in 557ed13.

test/Conversion/linalg2neura/expand-memref-copy.mlir
tancheng: Same fix applied here too. Fixed in 557ed13.

ShangkunLi (approved):

PR #345, opened 2026-09-01
Add template-guided mapping and Neura Python bindings

ShangkunLi:
## Summary

This PR adds the foundational support for authoring and compiling spatial templates in Neura.

It introduces a template-guided mapping strategy in which a frontend or expert-authored template fixes the spatial placement of operations, while the Neura backend remains responsible for validating the placement, scheduling the operations, routing values between tiles, and materializing the final mapped IR.

The PR also adds MLIR Python bindings for the Neura dialect, allowing Python-based frontends to construct Neura types, operations, and kernels without generating textual MLIR manually.

## Motivation

Neura’s existing heuristic mapper starts from an operation graph and automatically decides where operations should be placed. This is appropriate for general dataflow programs, but it does not provide a clean path for pre-designed spatial implementations such as operator templates.

For example, a template author may already know that the following dataflow should occupy three specific tiles:

```mlir
%lhs = "neura.constant"() <{value = 1 : i32}> {
  placement = {x = 0 : i32, y = 0 : i32}
} : () -> i32

%rhs = "neura.constant"() <{value = 2 : i32}> {
  placement = {x = 2 : i32, y = 0 : i32}
} : () -> i32

%result = "neura.add"(%lhs, %rhs) {
  placement = {x = 1 : i32, y = 0 : i32}
} : (i32, i32) -> i32
```

In this case, the compiler should preserve the authored placement rather than run a heuristic placement algorithm again. However, the template should not have to manually specify backend details such as link IDs, register allocation, routing paths, or execution time steps.

This PR establishes that separation:

```text
Template/frontend
    decides:
        operation → Tile placement

Neura backend
    decides:
        scheduling
        routing
        registers and links
        final mapping metadata
```

## Template-guided mapping

A new template mapping strategy is added to the existing accelerator mapping flow.

It can be selected with:

```bash
mlir-neura-opt input.mlir \
  --insert-data-mov \
  --map-to-accelerator="mapping-strategy=template mapping-mode=spatial-only"
```

For each materialized Neura operation, the template mapper:

1. Reads the operation’s `placement` attribute.
2. Resolves the corresponding physical Tile in the target architecture.
3. Verifies that the Tile supports the requested operation.
4. Preserves the authored spatial placement.
5. Searches for a legal execution time.
6. Uses the shared Neura routing infrastructure to route dependencies.
7. Emits the final `mapping_locs` and `mapping_info`.
8. Removes the frontend-only placement constraint after it has been materialized.

For example, an authored placement such as:

```mlir
placement = {x = 1 : i32, y = 0 : i32}
```

is materialized into backend mapping information similar to:

```mlir
mapping_locs = [{
  id = 1 : i32,
  resource = "tile",
  time_step = 1 : i32,
  x = 1 : i32,
  y = 0 : i32
}]
```

Data dependencies between placed operations are routed automatically and represented using link or register mapping locations:

```mlir
mapping_locs = [{
  id = 5 : i32,
  resource = "link",
  time_step = 0 : i32
}]
```

The resulting kernel also records how it was compiled:

```mlir
mapping_info = {
  compiled_ii = 1 : i32,
  mapping_mode = "spatial-only",
  mapping_strategy = "template",
  x_tiles = 4 : i32,
  y_tiles = 4 : i32
}
```

This keeps template authoring focused on the spatial computation while leaving physical routing and scheduling to the backend.

## Python bindings

This PR adds the C API registration and MLIR Python binding infrastructure required to use Neura from Python.

A Python frontend can now import and register the Neura dialect through the generated package:

```python
from neura_mlir.dialects import neura
from neura_mlir.ir import Context, Location

with Context(), Location.unknown():
    neura.register_dialect()

    # Neura types, operations, and kernels can be constructed here.
```

The bindings include:

- Neura dialect registration through the C API.
- Generated Python operation classes from `NeuraOps.td`.
- The custom Neura Python extension module.
- Neura types and operation builders.
- The required core MLIR Python modules and commonly used dialect bindings.

This provides the compiler interface needed by higher-level Python frontends such as Synapse. Those frontends can construct canonical MLIR objects directly instead of assembling Neura IR as text.

## Build and documentation updates

This PR also:

- Adds the Neura Python package to the CMake build.
- Adds the required C API and Python extension targets.
- Updates the CI workflow to build and test the Python bindings.
- Updates the README with Python binding build and usage instructions.
- Cleans up Clang build warnings without changing compiler behavior.


lib/NeuraDialect/Mapping/TemplateMapping/TemplateMapping.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+  if (!x_attr || !y_attr) {
+    op->emitError("Placement must contain integer x and y coordinates");
+    return nullptr;
+  }
+
+  int x = static_cast<int>(x_attr.getInt());
+  int y = static_cast<int>(y_attr.getInt());
+
+  for (Tile *tile : architecture.getAllTiles()) {
```
guosran: Could we use `Tile *Architecture::getTile(int x, int y)` instead?

ShangkunLi: Sure~ Template mapping now uses `Architecture::getTile(x, y)`.

guosran (approved):

PR #347, opened 2026-09-02
Add systolic template mapping with explicit LD/ST tiles

ShangkunLi:
## Summary

This PR adds template-guided mapping and code generation for systolic TileArray kernels.

It introduces:

- `neura.mac`, which produces an accumulated result and forwards its flowing input;
- explicit `neura.load` and `neura.store` Tiles for memory I/O;
- stationary data binding through kernel metadata;
- template mapping of computation and communication onto Tiles and links;
- general kernel code generation using the existing per-operation YAML/ASM path;
- tests for memory addressing, a MAC chain, a regular arithmetic kernel, and a 3×3 weight-stationary GEMM.

## Memory addressing

`neura.load` and `neura.store` support:

```mlir
// Dynamic address.
%v = "neura.load"(%addr) : (index) -> i32

// Runtime memref base with configured element offsets.
%v = "neura.load"(%A) {
  constants = array<i64: 0, 3, 6>
} : (memref<3x3xi32>) -> i32

// Configured absolute addresses.
%v = "neura.load"() {
  constants = array<i64: 32, 35, 38>
} : () -> i32
```

Store uses the same address forms, with the stored value as its first operand.

## Example: 3×3 GEMM on a 4×4 TileArray

The west column performs loads, the inner 3×3 Tiles perform MACs, and the south row performs stores:

```text
LD → MAC → MAC → MAC
       ↓     ↓     ↓
LD → MAC → MAC → MAC
       ↓     ↓     ↓
LD → MAC → MAC → MAC
       ↓     ↓     ↓
      ST    ST    ST
```

The stationary binding maps each MAC Tile `(x, y)` to:

```text
B[3 - y, x - 1]
```

```mlir
stationary = {
  kernel_input = 1 : i32,
  map = affine_map<(x, y) -> (3 - y, x - 1)>
}
```

A representative data path is:

```mlir
%a0 = "neura.load"(%a) {
  constants = array<i64: 0, 3, 6>,
  placement = {x = 0 : i32, y = 3 : i32}
} : (memref<3x3xi32>) -> i32

%m0:2 = "neura.mac"(%a0) {
  placement = {x = 1 : i32, y = 3 : i32}
} : (i32) -> (i32, i32)

%m1:2 = "neura.mac"(%m0#1) {
  placement = {x = 2 : i32, y = 3 : i32}
} : (i32) -> (i32, i32)
```

`%m0#0` is the partial sum flowing south, while `%m0#1` is the activation flowing east.

Code generation emits the corresponding `LOAD`, `MUL`, `MUL_ADD`, `


ShangkunLi:
@tancheng @guosran I have already renamed the `Port` -> `BoundaryPort`~

lib/NeuraDialect/Architecture/Architecture.cpp
```diff
@@ -567,6 +619,55 @@ void Architecture::applyLinkOverrides(
...
+//   - y increases from south to north.
+//
+// Each exposed boundary direction receives a distinct input port and output
+// port so that ingress and egress occupy independently modeled resources.
+void Architecture::initializeBoundaryPorts() {
+  int next_port_id = 0;
+
+  auto createPortPair = [&](Tile *tile, BoundaryPortDirection direction) {
+    for (BoundaryPortKind kind :
+         {BoundaryPortKind::Input, BoundaryPortKind::Output}) {
```
guosran: Do input & output ports always have to appear concurrently? In the gemm case, we only need input on the west and output on south.

lib/NeuraDialect/Mapping/TemplateMapping/TemplateMapping.cpp
```diff
@@ -40,6 +51,415 @@ Tile *getPlacedTile(Operation *op, const Architecture &architecture) {
...
+
+    mapping_state.reserveRoute(data_move.getOperation(), complete_path);
+    routed_inputs.push_back(data_move.getOperation());
+  }
+
+  return true;
+}
+
+// Routes kernel results from their producer Tiles to their assigned ports.
+bool routeKernelOutputPorts(KernelOp kernel,
```
guosran: `routeKernelOutputPorts` and `routeKernelInputPorts` have similar structures. Could we reuse the code?

guosran: `routeKernelOutputPorts` and `routeKernelInputPorts` have similar structures. Could we reuse some part of the code?

lib/NeuraDialect/Mapping/mapping_util.cpp
```diff
@@ -233,9 +236,10 @@ int mlir::neura::calculateResMii(Region &region,
 
   // Count all "compute" operations (non-terminators, non-block ops).
   region.walk([&](Operation *op) {
-    // Skips non-materialized ops.
-    if (isa<func::FuncOp>(op) ||
-        isa<neura::CtrlMovOp, neura::DataMovOp, neura::ReserveOp>(op)) {
+    // Only materialized operations consume Tile resources. Movement,
+    // recurrence placeholders, and region terminators do not contribute to
+    // ResMII.
+    if (isa<func::FuncOp>(op) || is_non_materialized(op)) {
```
guosran, line 262: Could we add FuncOp to `is_non_materialized`?

lib/NeuraDialect/Mapping/MappingState.cpp
```diff
@@ -920,6 +920,61 @@ void MappingState::encodeMappingState() {
...
+        kind_str = "boundary_port";
+        BoundaryPort *port = dyn_cast<BoundaryPort>(loc.resource);
+        assert(port && "Expected a BoundaryPort resource");
+
+        // A boundary port is attached to one boundary tile. The port direction
+        // describes the physical array boundary, while port kind describes
+        // whether data enters or leaves the TileArray through this resource.
+        Tile *tile = port->getTile();
+
+        int invalid_iterations = loc.time_step / II;
```
guosran: what is `invalid_iterations` and `index_per_ii`?

guosran: I mean why call it invalid?

lib/NeuraDialect/Transforms/PromoteInputArgToConstPass.cpp
```diff
@@ -47,6 +49,69 @@ LogicalResult promoteFunctionArgsToConstants(Region &region) {
   return success();
 }
 
+// Collects kernel input indices that are explicitly bound to physical input
+// ports by template metadata.
+//
+// These arguments must remain as block arguments because TemplateMapping uses
+// them as external streaming values injected through TileArray boundary ports.
+LogicalResult
+collectPortBoundInputIndices(neura::KernelOp kernel_op,
```
guosran: So we are only dealing with the inputs here? How do we stage the other matrix?

tancheng:
> a configured neura.mac that produces an accumulated result and forwards its flowing input;

Why I don't see `rhs const` in the following IR? "an accumulated result" should be "an accumulated result (after multiplication)"?

```
%m0:2 = "neura.mac"(%activation) {
  placement = {x = 0 : i32, y = 3 : i32}
} : (i32) -> (i32, i32)
```

> CGRA BoundaryPort resources for injecting and draining kernel values;

In our RTL/vectorCGRA, the systolic array matmul requires additional ld/st (one more row and one more column) to access data, right? Why you think your kernel can run in our RTL? And how do you verify its functionality? Does the generated yaml/instruction can run in Zeonica or VectorCGRA?

ShangkunLi:
> > a configured neura.mac that produces an accumulated result and forwards its flowing input;
> 
> Why I don't see `rhs const` in the following IR? "an accumulated result" should be "an accumulated result (after multiplication)"?
> 
> ```
> %m0:2 = "neura.mac"(%activation) {
>   placement = {x = 0 : i32, y = 3 : i32}
> } : (i32) -> (i32, i32)
> ```

The RHS is specified by the kernel’s stationary binding:
```
stationary = {
  kernel_input = 4 : i32,
  map = affine_map<(x, y) -> (3 - y, x)>
}
```
Kernel input 4 is B, so Tile (x,y) holds `B[3-y,x]`. Each MAC computes `input0 * B[3-y,x] + input1`, with an absent `input1` treated as zero, and forwards `input0` unchanged. For VectorCGRA, this binding should generate per-Tile `CMD_CONST` configuration, with `OPT_MUL_CONST` or `OPT_MUL_CONST_ADD`. That configuration generation is not implemented yet.


> > CGRA BoundaryPort resources for injecting and draining kernel values;
> 
> In our RTL/vectorCGRA, the systolic array matmul requires additional ld/st (one more row and one more column) to access data, right? Why you think your kernel can run in our RTL? And how do you verify its functionality? Does the generated yaml/instruction can run in Zeonica or VectorCGRA?

For memory I/O, I propose replacing the current `BoundaryPort` mapping with explicit `LD/ST` operations generated by the backend, following VectorCGRA’s systolic example. Then we can generate the yaml for VectorCGRA's configuration.

ShangkunLi:
Hi~ @tancheng  I have updated the implementation to use `LD/ST` Tiles instead of introducing `BoundaryPort` resources.

The example now computes a **3×3 GEMM on a 4×4 TileArray**: three `LD` Tiles on the west edge, nine MAC Tiles, and three `ST` Tiles on the south edge.

```text
LD → MAC → MAC → MAC
       ↓     ↓     ↓
LD → MAC → MAC → MAC
       ↓     ↓     ↓
LD → MAC → MAC → MAC
       ↓     ↓     ↓
      ST    ST    ST
```

Activations flow east; partial sums flow south. Each MAC computes `flowing_input * stationary_value + incoming_partial_sum` (zero when omitted) and forwards its flowing input separately.

For MAC Tile `(x, y)`, the stationary binding selects `B[3-y, x-1]`. B is a runtime kernel memref input, not a compile-time constant; its selected elements are installed in the Tiles’ constant queues before execution. `LD/ST` address queues similarly resolve element offsets against runtime SRAM base addresses, without separate address-computation Tiles.

I also extended `GenerateCodePass.cpp` to emit YAML/ASM using the existing instruction structure. The generated YAML was tested through the existing loader based on **`kernel-submit@366bdd7`**, with software adaptations for runtime bindings, address queues, and MAC input-lane selection.

The PyMTL RTL simulations passed, checking every output element against a reference GEMM.

tancheng:
> Hi~ @tancheng I have updated the implementation to use `LD/ST` Tiles instead of introducing `BoundaryPort` resources.
> 
> The example now computes a **3×3 GEMM on a 4×4 TileArray**: three `LD` Tiles on the west edge, nine MAC Tiles, and three `ST` Tiles on the south edge.
> 
> ```
> LD → MAC → MAC → MAC
>        ↓     ↓     ↓
> LD → MAC → MAC → MAC
>        ↓     ↓     ↓
> LD → MAC → MAC → MAC
>        ↓     ↓     ↓
>       ST    ST    ST
> ```
> 
> Activations flow east; partial sums flow south. Each MAC computes `flowing_input * stationary_value + incoming_partial_sum` (zero when omitted) and forwards its flowing input separately.
> 
> For MAC Tile `(x, y)`, the stationary binding selects `B[3-y, x-1]`. B is a runtime kernel memref input, not a compile-time constant; its selected elements are installed in the Tiles’ constant queues before execution. `LD/ST` address queues similarly resolve element offsets against runtime SRAM base addresses, without separate address-computation Tiles.
> 
> I also extended `GenerateCodePass.cpp` to emit YAML/ASM using the existing instruction structure. The generated YAML was tested through the existing loader based on **`kernel-submit@366bdd7`**, with software adaptations for runtime bindings, address queues, and MAC input-lane selection.
> 
> The PyMTL RTL simulations passed, checking every output element against a reference GEMM.

That's awesome, is the pymtl rtl simulation test is also in this repo or anywhere else?

For the `neura.mac`, I still want to discuss about it (maybe in our weekly meeting). What is the format of `neura.add` with rhs constant? Can the `neura.mac` be aligned with `neura.add`? I understand for weight-stationary, if rhs is also an activation, it is unknown during compilation, which actually can be viewed as rhs-stationary, but we indeed need to align their representation, as the data is passed in via CMD_CONST, WDYT?

ShangkunLi:
> > Hi~ @tancheng I have updated the implementation to use `LD/ST` Tiles instead of introducing `BoundaryPort` resources.
> > The example now computes a **3×3 GEMM on a 4×4 TileArray**: three `LD` Tiles on the west edge, nine MAC Tiles, and three `ST` Tiles on the south edge.
> > ```
> > LD → MAC → MAC → MAC
> >        ↓     ↓     ↓
> > LD → MAC → MAC → MAC
> >        ↓     ↓     ↓
> > LD → MAC → MAC → MAC
> >        ↓     ↓     ↓
> >       ST    ST    ST
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Activations flow east; partial sums flow south. Each MAC computes `flowing_input * stationary_value + incoming_partial_sum` (zero when omitted) and forwards its flowing input separately.
> > For MAC Tile `(x, y)`, the stationary binding selects `B[3-y, x-1]`. B is a runtime kernel memref input, not a compile-time constant; its selected elements are installed in the Tiles’ constant queues before execution. `LD/ST` address queues similarly resolve element offsets against runtime SRAM base addresses, without separate address-computation Tiles.
> > I also extended `GenerateCodePass.cpp` to emit YAML/ASM using the existing instruction structure. The generated YAML was tested through the existing loader based on **`kernel-submit@366bdd7`**, with software adaptations for runtime bindings, address queues, and MAC input-lane selection.
> > The PyMTL RTL simulations passed, checking every output element against a reference GEMM.
> 
> That's awesome, is the pymtl rtl simulation test is also in this repo or anywhere else?

The PyMTL test is currently local. It uses @Jackcuii's [kernel-submit@366bdd7](https://github.com/tancheng/VectorCGRA/tree/kernel-submit) branch plus our local yaml loader adaptations. Since that branch is not yet merged into main, I haven’t pushed the integration test or loader changes online.

> For the `neura.mac`, I still want to discuss about it (maybe in our weekly meeting). What is the format of `neura.add` with rhs constant? Can the `neura.mac` be aligned with `neura.add`? I understand for weight-stationary, if rhs is also an activation, it is unknown during compilation, which actually can be viewed as rhs-stationary, but we indeed need to align their representation, as the data is passed in via CMD_CONST, WDYT?

A `MAC` mathematically has three operands—two multiplicands and one addend:

```text
result = a * b + c
```

Our current `neura.mac` exposes one multiplicand and the incoming partial sum as dynamic inputs:

```tablegen
let arguments = (
  ins AnyType:$input0,
      Optional<AnyType>:$input1
);
let results = (outs AnyType:$result, AnyType:$forwarded);
```

The other multiplicand, `s`, is configured through the enclosing kernel and loaded via `CMD_CONST`:

```text
result    = input0 * s + input1   // input1 defaults to zero
forwarded = input0
```

This supports **weight-stationary** when `s` is a weight, and **input-stationary** when `s` is an activation.

**Output-stationary** instead takes two flowing multiplicands and retains the partial sum locally:

```text
local_accumulator += input0 * input1
```

We do not support this in our RTL yet. The existing [SeqMulAdderRTL](https://github.com/tancheng/VectorCGRA/blob/366bdd71666b6c0b1a1848a21d1196dc5bd9d1e6/fu/basic/TwoSeqCombo.py)/[TwoSeqCombo](https://github.com/tancheng/VectorCGRA/blob/366bdd71666b6c0b1a1848a21d1196dc5bd9d1e6/fu/basic/TwoSeqCombo.py) takes the addend from an external input; it does not maintain an internal accumulator across firings. Implementing OS through Tile-register feedback would require additional initialization, feedback, and result-draining support.

tancheng:
Why `forwarded` is required in `let results = (outs AnyType:$result, AnyType:$forwarded);`?

Can't we just use dependency to represent it?

So the question is why we need to have neura IR to represent matmul? Can't it be a black box from neura perspective? e.g., `neura.custom_kernel`.

ShangkunLi:
> Why `forwarded` is required in `let results = (outs AnyType:$result, AnyType:$forwarded);`?
> 
> Can't we just use dependency to represent it?
> 
> So the question is why we need to have neura IR to represent matmul? Can't it be a black box from neura perspective? e.g., `neura.custom_kernel`.

`forwarded` is exactly a dependency. Without it:

```text
p0 = mac(a0, ...) @ Tile(0, y)
p1 = mac(a0, ...) @ Tile(1, y)
p2 = mac(a0, ...) @ Tile(2, y)
```

All MACs use the same source. This does not specify that `a0` flows left-to-right through neighboring Tiles.

With `forwarded`:

```text
p0, a1 = mac(a0, ...) @ Tile(0, y)
p1, a2 = mac(a1, ...) @ Tile(1, y)
p2, a3 = mac(a2, ...) @ Tile(2, y)
```

Now the dependencies explicitly describe `Tile(0,y) → Tile(1,y) → Tile(2,y)`. Neura maps these transfers onto the corresponding links.

For `neura.custom_kernel`, the question is: how does a user implement a new custom kernel?

1. [Synapse](https://github.com/coredac/synapse) lets them describe its spatial computation and data movement.
2. [Neura](https://github.com/coredac/neura) assigns Tile, link, and register resources and generates YAML.
3. Without this compilation flow, users would have to write and debug hardware YAML themselves for every new kernel?

A black-box call can invoke an already compiled kernel, which is implemented as pattern matching and rewriting in `synapse` now, but we still need this flow to define & build that custom kernel.

tancheng:
Then shouldn't it be `%m0:2 = synapse.mac` rather than `%m0:2 = neura.mac`?

ShangkunLi:
> Then shouldn't it be `%m0:2 = synapse.mac` rather than `%m0:2 = neura.mac`?

At the Python programming level, it is `synl.mac`. After lowering, it becomes `neura.mac`, which Neura maps to the hardware:

```text
synl.mac → neura.mac → Tile/FU/crossbar configuration
```

The distinction is that `neura.mac` describes a **configured Tile’s computation and forwarding**, not just the arithmetic FU. Its `result` comes from the FU, while `forwarded` uses the routing crossbar.

Those dependencies need to remain in Neura IR so the mapper and codegen can allocate the links/registers and configure the forwarding path.

tancheng:
Thanks @ShangkunLi, let's continue the discussion tmr about the `rhs_const`, and `forward`.

ShangkunLi:
> Thanks @ShangkunLi, let's continue the discussion tmr about the `rhs_const`, and `forward`.

@tancheng I pushed the latest update.

This commit mainly does two things:

1. It refines `neura.load/store` to represent dynamic addresses, memref-base-plus-offsets, and configured addresses directly.
2. It removes the GEMM-specific codegen path. Configured LD/ST/MAC operands now extend the existing per-operation code generator, so regular mapped kernels use the same path.

For example:

```mlir
%a = "neura.load"(%A) {
  constants = array<i64: 0, 3, 6>
} : (memref<3x3xi32>) -> i32
```

This means that successive executions read:

```text
base(A) + 0
base(A) + 3
base(A) + 6
```

The 3×3 GEMM on a 4×4 TileArray now lowers to three LOADs, nine MACs, and three STOREs, then generates the corresponding YAML/ASM.

I also added codegen coverage for a regular `constant + add` kernel to confirm that the code generator is not limited to GEMM. The generated GEMM YAML was validated locally against the VectorCGRA PyMTL RTL using `kernel-submit@366bdd7` with local loader adaptations.

lib/NeuraDialect/Architecture/Architecture.cpp
ShangkunLi: I've already removed the `port` in the latest commits.

tancheng:
> The generated GEMM YAML was validated locally against the VectorCGRA PyMTL RTL using kernel-submit@366bdd7 with local loader adaptations.

The VectorCGRA master can merge @Jackcuii's fix (https://github.com/tancheng/VectorCGRA/pull/336) (excluding the register cluster hack) soon. Can we later update https://github.com/tancheng/CGRA-Flow to include both latest neura and VectorCGRA for the verification? (I mean file an issue in CGRA-Flow, and provide the test with help from Jia or anyone else, WDYT? plz reply the email if you think it is a good plan, you can let both of them work on different tests -- I like this idea more :-) So the email can CC or @ the other person who is working on the interpreter. You can also share your changes in kernel-submit@366bdd7 to give them an example).

tancheng (approved):

PR #322, opened 2026-06-10
Handle mapped form of neura.add in interpreter

TimJZ:
## Background

The Neura spatial mapper (\`--map-to-accelerator\`) transforms pre-mapping IR into a tile-assigned, physically-routed form. As part of this transformation, compile-time constant operands are folded from SSA values into operation attributes. Specifically, an instruction like:

\`\`\`mlir
%res = "neura.add"(%lhs, %const_1) : (!neura.data<i64, i1>, !neura.data<i64, i1>) -> !neura.data<i64, i1>
\`\`\`

becomes:

\`\`\`mlir
%res = "neura.add"(%lhs) {rhs_value = 1 : i64} : (!neura.data<i64, i1>) -> !neura.data<i64, i1>
\`\`\`

This single-SSA-operand form is the intended post-mapping representation — the \`\$rhs\` field in \`Neura_AddOp\` is defined as \`Optional<AnyType>\` in \`NeuraOps.td\` precisely to support it. However, \`handleAddOp\` in the interpreter unconditionally rejected any op with fewer than two SSA operands, making it impossible to interpret any fully mapped IR that contains \`neura.add\`.

This was observed when running \`neura-interpreter\` on the ReLU e2e mapped IR (\`test/e2e/relu/Output/relu_kernel.mlir.tmp-mapping.mlir\`), where the loop index increment \`i + 1\` maps to:

\`\`\`mlir
%23 = "neura.add"(%22) {rhs_value = 1 : i64} : (!neura.data<i64, i1>) -> !neura.data<i64, i1>
\`\`\`

## PR Summary

**Interpreter**
- Extend \`handleAddOp\` to handle the single-operand mapped form: detect \`numOperands == 1\`, read the constant from the \`rhs_value\` \`IntegerAttr\`, and synthesize a \`PredicatedData\` with \`predicate = true\` (compile-time constants are always valid)
- Retain the existing two-operand path unchanged so pre-mapping IR continues to work
- Add a zero-operand error path with a clear diagnostic

**Tests**
- Add Test 5 to \`test/neura/interpreter/basic_operation/add.mlir\`: uses \`neura.grant_once\` to produce a typed \`!neura.data<i64, i1>\` value, then applies \`neura.add\` with \`rhs_value = 5\` and verifies the output is \`15.000000\`
- All existing tests (Tests 1–4, two-operand pre-mapping form) continue to pass

PR #350, opened 2026-09-08
Handle mapped form of neura.icmp in interpreter

TimJZ:
## Background

Same root cause as #322's neura.add fix. After --map-to-accelerator,
FoldConstantPass folds compile-time constant operands into a rhs_value
attribute:

    %res = "neura.icmp"(%lhs) {cmpType = "sge", rhs_value = 0 : i32}
             : (!neura.data<i32, i1>) -> !neura.data<i1, i1>

handleICmpOp unconditionally rejected any op with fewer than two SSA
operands. This blocks interpreting relu_neura_mapped.mlir, which uses
this exact form twice (the ReLU sign check and the loop-bound check).

## PR Summary

**Interpreter**
- Extend handleICmpOp to handle the single-operand mapped form, same
  pattern as the merged handleAddOp fix
- Retain the existing two-operand path unchanged
- Follow-up cleanup (per @ShangkunLi's suggestion on #322): extract the
  shared "single-operand + rhs_value attribute" resolution logic out of
  handleAddOp and handleICmpOp into one resolveBinaryOperands helper, so
  any future binary op needing this treatment doesn't have to copy-paste
  it again. Pure refactor, no behavior change.

**Tests**
- Add two cases to icmp.mlir modeled on the real mapped-IR usage:
  - test_icmp_mapped_form_true — grant_once constant 5, sge 0 → 1
  - test_icmp_mapped_form_false — grant_once constant -5, sge 0 → 0
- All 18 existing two-operand cases (eq, ne, slt, sle, sgt, sge, ult,
  ule, ugt, uge) continue to pass, confirming no regression on the
  pre-mapping path
- Full suite regression: 96/97 both before and after the cleanup
  commit (one pre-existing, unrelated local-environment failure —
  missing torch module)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/test/neura/interpreter/basic_operation/icmp.mlir
+++ b/test/neura/interpreter/basic_operation/icmp.mlir
@@ -169,6 +169,24 @@ func.func @test_icmp_ugt_false() -> i1 {
   return %ugt : i1
 }
 
+// ====== Mapped form — single SSA operand with rhs_value attribute ======
+// After --map-to-accelerator, compile-time constant operands are folded into
+// rhs_value attributes (e.g. "neura.icmp"(%x) {cmpType = "sge", rhs_value = 0 : i32}).
+// Before this fix, the interpreter rejected any op with numOperands < 2.
+func.func @test_icmp_mapped_form_true() -> !neura.data<i1, i1> {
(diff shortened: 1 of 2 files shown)


tancheng (approved):

tancheng:
LGTM, feel free to merge the PR, thanks~! @TimJZ

